PR #16, opened 2019-06-06
Simplify message passing via Comlink

e111077:
Moving code output from src -> src to src -> lib stuff made the diff weird: [this commit](https://github.com/PolymerLabs/code-sample-editor/commit/ec7a1ee4574624ca41f7d7148624b9865e1e5199) has the beef of the changes; the others are moving files around.

# In this PR
* Removed a lot of my messaging code
* Replaced a lot of it with [comlink](https://github.com/GoogleChromeLabs/comlink)
  * Did not replace initial MessageChannel handhsake
  * MessageChannels needed because SW needs to know which client to respond to
  * Cannot use `Clients.get(fetchEvent.clientId)` because that returns the iframe
* Moved files so that output JS files is not alongside TS files in `/src` but rather `/lib`

# Things to note / Requesting feedback
* ~`importScripts('./lib/comlink.js')` in `sw.ts`~
  * ~Looking for better solution there as modules don't get exist in SWs~
  * ~Cannot use `import.meta`~
  * ~file copied via `postinstall` script~
* ~`SwControllerInterface` in `sw.ts` is extraneous:~
  * ~necessary because all exports had to be compiled out~
  * ~`typeof SWController` would cause declaration file compilation fail~

**Solved with bundling with rollup -> iife**

# Related Future work
Use comlink to push TSC work to web Worker. If we use monaco in an iframe, this can simplify communication across the iframe border.

Try to move to stateless (memory-efficient) serviceworker either stackblitz's localservice or use comlink to reply directly to fetch requests


src/lib/code-sample-editor.ts
```diff
@@ -25,44 +27,24 @@ export class CodeSampleEditor extends LitElement {
...
-        port.removeEventListener('message', this.onSwMessage);
-        clearSession(this.sessionId, port);
-      });
 
-      this.swSetup = setUpServiceWorker(this.sandboxScope);
-
-      this.swPortEstablished = this.swSetup
-          .then((response) => {
-        const sw = response[0];
+      this.remoteSw = this.remoteSw.then(sw => {
```
usergenic: <🏌️‍♂️>
```ts
this.remoteSw = this.remoteSw
  .then(sw => sw && sw.clearContents(this.sessionId))
  .then(async _ => await setUpServiceWorker(this.sandboxScope));
```
</🏌️‍♂️>

src/sw.ts
```diff
@@ -0,0 +1,130 @@
...
+
+importScripts('./lib/comlink.js');
+
+const swScope = self as ServiceWorkerGlobalScope;
+const Comlink = (swScope as any).Comlink as typeof CL;
+
+const MESSAGE_TYPES = {
+  ESTABLISH_HANDSHAKE: "ESTABLISH_HANDSHAKE",
+  HANDSHAKE_RECEIVED: "HANDSHAKE_RECEIVED",
+} as unknown as typeof MESSAGE_TYPES_;
```
usergenic: It feels unfair that you should have to type these out again like this.

I don't know off-hand if there's a better solution, but I thought you should know I think that's unfair.

usergenic: I had a similar issue in https://github.com/Polymer/koa-node-resolve/blob/master/src/support/logger.ts with LogLevel.

src/sw.ts
```diff
@@ -0,0 +1,130 @@
...
+  content: string,
+  init: ResponseInit
+}
+
+let fileResponseMap = new Map<string, Map<string, ResponseParams>>();
+
+const endWithSlash = (str:string) => {
+  const endsWithSlash = str[str.length - 1] === '/';
+  return endsWithSlash ? str : `${str}/`
+}
```
usergenic: <🏌️‍♀️>
```ts
const endsWithSlash = (str: string) => str.replace(/\/?$/, '/');
```
</🏌️‍♀️>

src/sw.ts
e111077: It would be great if sw modules were here. Perhaps I should compile the SW into a single file?

src/sw.ts
e111077: I can hear the _thwuck_ of hitting the golfball

src/lib/code-sample-editor.ts
e111077: had to add async or else TS doesn't like returning `null | PromiseLike<void>`

PR #23, opened 2020-07-24
Split into multiple TypeScript projects

justinfagnani:
This creates `/src/service-worker` and `/src/shared` folders with their own tsconfig files so that they have the correct lib settings (no DOM lib). There will also be a `/src/typescript-worker` folder in the future.

util.ts was split into two files and the functions moved around solely based on what types they needed, not by who calls them. This is a temporary organization since I'll move things further in a following PR.

This PR doesn't change any of the service worker or component code, except for some compiler warning fixes.


src/lib/code-sample-editor.ts
```diff
@@ -8,20 +8,23 @@ import {
   query,
   queryAll,
 } from 'lit-element';
-import {until} from 'lit-html/directives/until';
-import {FileRecord, CodeSampleEditorTextarea, RemoteSw} from './types';
-import {EMPTY_INDEX} from './constants';
+import {until} from 'lit-html/directives/until.js';
```
e111077, line 11: Google style guide says leave this off. Do we care? will this ever be used internally? (IK it'd be third_party, but being able to work on it and work towards TS readability is appreciated)

src/shared/types.ts
```diff
@@ -0,0 +1,83 @@
...
+
+declare global {
+  interface Window {
+    __CodeEditorSessions?: Set<string>;
+  }
+  interface ImportMeta {
+    url: string;
+  }
+
+  // interface FetchEvent extends Event {
```
e111077, line 12: why are these commented and not deleted?

src/lib/code-sample-editor.ts
justinfagnani: Google style guide is workable internally where the files are never directly loaded into browsers, but the only way to get these to work natively in browsers right from the tsc output is to include the file extensions. You may see things working because es-dev-server will do the node resolution even for relative paths, but that's relying on a specific and non-standard tool. If you happen to use extension-less imports in MWC, I'd change those for maximum compatibility externally.

src/shared/types.ts
justinfagnani: oops

justinfagnani: fixed

PR #24, opened 2020-07-24
Update service worker logic, add MWC components

justinfagnani:
This makes the service worker more reliable, and works even with devtools (and its reload service worker option) is open. I think there's on remaining case where a page was open for a long time where the service worker can be unregistered and the connection to the page lost. Still working on that one.

This also adds some MWC components. It was a little hard to separate the changes. I did remove the CodeMirror component, and TypeScript worker to save for the next PRs.


demo/index.html
```diff
@@ -2,12 +2,20 @@
...
+  </style>
 </head>
+
 <body>
-  <code-sample-editor project-path="./templates"></code-sample-editor>
-  <code-sample-editor project-path="./templates"></code-sample-editor>
-  <code-sample-editor project-path="./templates" sandbox-scope="myScope"></code-sample-editor>
+  <!-- <code-sample-editor project-src="./typescript/project.json"></code-sample-editor> -->
+  <code-sample-editor project-src="./project-1/project.json"></code-sample-editor>
+  <code-sample-editor project-src="./mwc-button/project.json"></code-sample-editor>
```
justinfagnani, line 18: I changed the attribute name to end in `src` so it has similar semantics to built-in `src` attributes, and the value to be a file so we don't have a filename we look for. From the config, demo files just follow relative paths.

src/lib/code-sample-editor-preview.ts
```diff
@@ -0,0 +1,110 @@
...
+  css,
+  property,
+  query,
+} from 'lit-element';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import '@material/mwc-icon-button';
+import '@material/mwc-textfield';
+
+@customElement('code-sample-editor-preview')
+export class CodeSampleEditorPreviewElement extends LitElement {
```
justinfagnani, line 28: I should add a doc comment. This is a component responsible for setting up the iframe, and has a URL bar-like header with the file name and a reload button. It's intended to evolve to a more full-featured preview component, maybe with a console log viewer, drop-down to select different html files to view, etc.

src/service-worker/service-worker.ts
```diff
@@ -1,37 +1,57 @@
...
-  const ports = e.ports;
-  if (ports && ports[0]) {
-    const port = ports[0];
-    port.start();
-    const handshakeReceivedMessage: Message = {
-      type: MESSAGE_TYPES.HANDSHAKE_RECEIVED,
-    };
-    expose(SwController, port);
-    port.postMessage(handshakeReceivedMessage);
+const fileAPIs = new Map<string, FileAPI>();
```
justinfagnani, line 36: I should add a comment. The file cache was replaced with an API proxied from the page that provides one file at a time. This allows the page to only send the files that are requested, and also wait for compilation, when that's added.

src/service-worker/service-worker.ts
```diff
@@ -1,37 +1,57 @@
...
-let fileResponseMap = new Map<string, Map<string, ResponseParams>>();
-
 const onFetch = (e: FetchEvent) => {
-  if (!e.request || !e.request.url || !e.respondWith) {
-    return;
-  }
   const url = new URL(e.request.url);
   const href = url.href;
-  const scope = endWithSlash(swScope.registration.scope);
+  const scope = endWithSlash(self.registration.scope);
```
e111077, line 54: Super congrats on getting TS to accept this as a SW

src/service-worker/service-worker.ts
```diff
@@ -1,37 +1,57 @@
...
+
+declare var self: ServiceWorkerGlobalScope;
+
+/**
+ * API exposed to the UI thread via Comlink. The static methods on this class
+ * become instance methods on SwControllerAPI.
+ */
+const workerAPI: ServiceWorkerAPI = {
+  setFileAPI(fileAPI: FileAPI, sessionID: string) {
+    fileAPIs.set(sessionID, fileAPI);
```
e111077, line 32: Ugh this is so clean I'm jealous

e111077: It's kinda amazing how simple and powerful SWs are yet such a PiTA to use when it comes to the nitty semantics

src/service-worker/service-worker.ts
```diff
@@ -1,37 +1,57 @@
...
-    const handshakeReceivedMessage: Message = {
-      type: MESSAGE_TYPES.HANDSHAKE_RECEIVED,
-    };
-    expose(SwController, port);
-    port.postMessage(handshakeReceivedMessage);
+const fileAPIs = new Map<string, FileAPI>();
+const getFile = async (e: FetchEvent, path: string, sessionId: string) => {
+  const fileAPI = fileAPIs.get(sessionId);
+  if (fileAPI) {
+    const file = await fileAPI.getFile(path);
```
e111077, line 40: How is performance on this? I remember creating the file record cache simply because you said it was taking too long to reload fetching from the client every time

src/lib/code-sample-editor-preview.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    iframe {
+      flex: 1;
+      width: 100%;
+      border: none;
+    }
+  `;
+
+  @property()
+  src?: string;
```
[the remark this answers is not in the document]
justinfagnani: I need undefined here to represent the uninitialized case and to work with `ifDefined` before. I much prefer using `undefined` as the uninitialized value for JS objects, when not emulating DOM APIs, because it's the default value for a property any way. This shouldn't cause any problems with debugging. What have you seen?

src/lib/code-sample-editor-preview.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  @property()
+  location?: string;
+
+  @query('iframe')
+  private _iframe!: HTMLIFrameElement;
```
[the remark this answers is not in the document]
justinfagnani: I strongly disagree with Google style here. We can't force our users to use TypeScript, which will tell them that this is a private API. If they're using JavaScript they should have the standard indicators that this is protected or private, like the `_` prefix. This is what we do for all of our public libraries.

src/lib/code-sample-editor-preview.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+  `;
+
+  @property()
+  src?: string;
+
+  @property()
+  location?: string;
+
+  @query('iframe')
```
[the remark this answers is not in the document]
justinfagnani: This is called only on user interaction, so the perf is fine and I'd prefer non-caching to extra memory for the cached value.

src/service-worker/service-worker.ts
justinfagnani: took a bit!

src/service-worker/service-worker.ts
justinfagnani: Thanks! And ell me about it. I just got the last flakiness issue ironed out too I think.

src/service-worker/service-worker.ts
justinfagnani: Hmm... I don't remember that, maybe it had to do with compilation? If we were going to do compilation in the service worker, we'd need all the files sent over. But I put compilation in a separate worker so that it wouldn't tie up the service worker thread and it could fetch other files for other projects in parallel. The perf here is fine, copying strings is fast.

justinfagnani:
I think I figured out the last bit of flakiness. With devtools closed, or the "update service workers on reload" option off, there was a case where the preview iframe would request a URL before the FileAPI had been sent from the component to the service worker, resulting in a 404. I solved this by making `_files` and `_serviceWorkerAPI` reactive properties, and only generating the `src` property for the preview when those two are defined. This way we have a FileAPI and files to serve when the preview boots up.

@e111077 PTAL

src/lib/code-sample-editor-preview.ts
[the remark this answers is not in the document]
e111077: often issues with components are startup states and I've often run into issues where debugging we don't know what part of the render cycle it was in because the default value was `undefined` rather than our initialization of `null`. IMO `ifDefined` needs a parent function / directive that can take a comparison basis. See textfield:

https://github.com/material-components/material-components-web-components/blob/master/packages/textfield/src/mwc-textfield-base.ts#L317

Also counterpoint, `iframe.src` is an empty string if undefined and so is `input.value`

src/lib/code-sample-editor-preview.ts
```diff
@@ -0,0 +1,122 @@
...
+   * The string to display in the location bar.
+   */
+  @property()
+  location?: string;
+
+  @query('iframe')
+  private _iframe!: HTMLIFrameElement;
+
+  render() {
+    console.log('preview src', this.src);
```
e111077, line 99: cleanup

src/lib/code-sample-editor-preview.ts
```diff
@@ -99,8 +110,9 @@ export class CodeSampleEditorPreviewElement extends LitElement {
   }
 
   reload() {
-    if (this._iframe.contentWindow) {
-      this._iframe.contentWindow.location.reload();
+    const iframe = this._iframe;
+    if (iframe.contentWindow) {
```
e111077, line 114: ripe for optional chaining if you're up for it

src/lib/code-sample-editor.ts
```diff
@@ -211,12 +224,12 @@ export class CodeSampleEditor extends LitElement {
...
         </mwc-tab-bar>
         <textarea
-          .value=${this._currentFile && this._currentFile.content || ''}
+          .value=${(this._currentFile && this._currentFile.content) || ''}
           @change=${this._onEdit}
         ></textarea>
       </div>
       <code-sample-editor-preview
-        src="${indexUrl.href}"
+        .src="${this._previewSrc}"
```
e111077, line 232: get rid of quotes for consistency

src/service-worker/service-worker.ts
```diff
@@ -18,53 +18,61 @@ import {
...
+type SessionID = string;
+
 /**
  * API exposed to the UI thread via Comlink. The static methods on this class
  * become instance methods on SwControllerAPI.
  */
 const workerAPI: ServiceWorkerAPI = {
-  setFileAPI(fileAPI: FileAPI, sessionID: string) {
+  setFileAPI(fileAPI: FileAPI, sessionID: SessionID) {
+    console.log(`Setting FileAPI for session ${sessionID}`);
```
e111077, line 34: cleanup

src/service-worker/service-worker.ts
```diff
@@ -18,53 +18,61 @@ import {
...
-  const href = url.href;
+  const url = e.request.url;
   const scope = endWithSlash(self.registration.scope);
-  if (href.startsWith(scope)) {
-    const fullPath = href.substring(scope.length);
-    const fullPathParts = fullPath.split('/');
-    const sessionId = fullPathParts.shift();
+  if (url.startsWith(scope)) {
+    const fullPath = url.substring(scope.length);
+    const slashIndex = fullPath.indexOf('/');
```
e111077, line 65: perhaps a small comment above about how the url schema should look? Hard to keep track with the slices and index manipulation

src/lib/code-sample-editor-preview.ts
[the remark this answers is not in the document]
justinfagnani: Since defaults are set in the constructor, independent of any LitElement lifecycle, I don't think there should be a question of when in the render cycle you are. undefined doesn't give you any different information than null, since the constructor has to run.

src/lib/code-sample-editor.ts
```diff
@@ -1,201 +1,365 @@
...
+   */
+  // TODO: generate this?
+  @property({attribute: 'sandbox-scope'})
+  sandboxScope = 'code-sample-editor-projects';
+
+  // computed from this.sandboxScope
+  _scopeUrl!: string;
+
+  @property({type: Boolean})
+  enableAddFile = false;
```
aomarks, line 155: Document

src/lib/code-sample-editor.ts
```diff
@@ -1,201 +1,365 @@
...
-    const isNewScope = this.lastSandboxScope !== this.sandboxScope;
-
-    if (isNewScope) {
-      this.lastSandboxScope = this.sandboxScope;
-      this.remoteSw = connectToServiceWorker(
-        this.remoteSw,
-        this.sessionId,
-        this.sandboxScope
-      );
+  private async _startWorkders() {
```
aomarks, line 282: s/Workders/Workers

aomarks (approved):

package.json
```diff
@@ -18,14 +18,19 @@
   "author": "The Polymer Authors",
   "license": "BSD-3-Clause",
   "devDependencies": {
-    "es-dev-server": "^1.56.1",
+    "es-dev-server": "~1.56.1",
```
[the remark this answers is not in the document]
justinfagnani: Newer versions were injecting a script into .html files, which was visible in the samples. This might have been fixed with a flag since then.

src/lib/code-sample-editor.ts
justinfagnani: done

PR #25, opened 2020-07-24
Add <codemirror-editor>

justinfagnani:


src/lib/codemirror-editor.ts
```diff
@@ -0,0 +1,145 @@
...
+   */
+  @property()
+  type: 'js' | 'html' | 'css' | undefined;
+
+  render() {
+    return html`${this._editorView?.dom}`;
+  }
+
+  update(changedProperties: PropertyValues) {
+    if (changedProperties.has('value') || changedProperties.has('src')) {
```
e111077, line 92: Would there be any practical use as to check that this.value and this.src are not being unset?

justinfagnani: It's fine if they're unset, we just need to make a new view if the content is changed from the outside.

Actually, we might not always need to, but I'm still learning the CodeMirror6 API. This works for the limited use case here.

src/lib/codemirror-editor.ts
```diff
@@ -0,0 +1,145 @@
+/**
+ * @license
+ * Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
```
e111077, line 3: 2020

aomarks (approved):

src/lib/codemirror-editor.ts
```diff
@@ -0,0 +1,145 @@
...
+      }),
+      root: this.shadowRoot!,
+    });
+    this._editorView = view;
+    this.requestUpdate();
+  }
+
+  private _getLanguagePlugins() {
+    switch (this.type) {
+      case 'js':
```
e111077, line 136: is TS not a built in one or is is jslang?

e111077 (changes requested):
nits and questions

e111077 (approved):
preapproval

src/lib/codemirror-editor.ts
justinfagnani: done

src/lib/codemirror-editor.ts
justinfagnani: I think TS just started landing as a "variant" of JS. I'll look into it soon.

PR #26, opened 2020-07-25
Add TypeScript compilation

justinfagnani:
Adds a TypeScript worker to compile sample source files. There is one worker per instance of code-sample-editor.

The loading and compiler setup is pretty basic still, but enough to run TypeScript samples. More work will have to be done to get full type-checking: recursively loading _all_ files, loading the correct lib files, and doing a better job of loading .d.ts files for .js dependencies.

### DIFF
--- a/src/lib/code-sample-editor.ts
+++ b/src/lib/code-sample-editor.ts
@@ -35,6 +35,7 @@ import {
   ProjectManifest,
   ESTABLISH_HANDSHAKE,
   HANDSHAKE_RECEIVED,
+  TypeScriptWorkerAPI,
 } from '../shared/worker-api.js';
 import {getRandomString, endWithSlash} from '../shared/util.js';
 import {CodeSampleEditorPreviewElement} from './code-sample-editor-preview.js';
@@ -63,6 +64,10 @@ const generateUniqueSessionId = (): string => {
 
 const serviceWorkerScriptUrl = new URL(
   '../../service-worker.js',
+  import.meta.url
+);
+const typescriptWorkerScriptUrl = new URL(
+  '../../typescript-worker.js',
   import.meta.url
 );
 
@@ -190,6 +195,11 @@ export class CodeSampleEditor extends LitElement {
 
   @internalProperty()
   private _serviceWorkerAPI?: Remote<ServiceWorkerAPI>;
+  private _typescriptWorkerAPI?: Remote<TypeScriptWorkerAPI>;
+  private _compiledFilesPromise = Promise.resolve<
+    Map<string, string> | undefined
+  >(undefined);
+  private _compiledFiles?: Map<string, string>;
 
   private get _previewSrc() {
     // Make sure that we've connected to the Service Worker and loaded the
@@ -278,7 +288,7 @@ export class CodeSampleEditor extends LitElement {
         };
       })
     );
-    // TODO (justinfagnani): compile project here
+    this._compileProject();
     this._currentFileIndex = 0;
     // TODO(justinfagnani): whyyyy?
     await this._tabBar.updateComplete;
@@ -288,9 +298,18 @@ export class CodeSampleEditor extends LitElement {
 
   private async _startWorkers() {
     await Promise.all([
-      // TODO (justinfagnani): this._startTypeScriptWorker(),
+      this._startTypeScriptWorker(),
       this._installServiceWorker(),
     ]);
+  }
+
+  private async _startTypeScriptWorker() {
+    if (this._typescriptWorkerAPI === undefined) {
+      const worker = new Worker(typescriptWorkerScriptUrl);
+      this._typescriptWorkerAPI = wrap<TypeScriptWorkerAPI>(worker);
+    } else {
+      console.debug('typescript-worker already started');
+    }
   }
 
   private async _installServiceWorker() {
@@ -331,12 +350,7 @@ export class CodeSampleEditor extends LitElement {
           this._serviceWorkerAPI = wrap<ServiceWorkerAPI>(port1);
           this._serviceWorkerAPI.setFileAPI(
             proxy({
-              getFile: async (
-                name: string
-              ): Promise<SampleFile | undefined> => {
-                // TODO (justinfagnani): compile files here
-                return this._files?.find((f) => f.name === name);
-              },
+              getFile: (name: string) => this._getFile(name),
             }),
             this._sessionId
           );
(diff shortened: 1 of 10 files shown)


src/lib/code-sample-editor.ts
```diff
@@ -65,6 +66,10 @@ const serviceWorkerScriptUrl = new URL(
   '../../service-worker.js',
   import.meta.url
 );
+const typescriptWorkerScriptUrl = new URL(
+  '../../typescript-worker.js',
+  import.meta.url
```
e111077, line 71: only problem I have with workers is that they come with lots of caveats in bundling. Perhaps one day it would be nice to have a more-modular approach e.g. give link to worker or something? or is that too much of a security risk?

src/lib/code-sample-editor.ts
```diff
@@ -274,7 +284,7 @@ export class CodeSampleEditor extends LitElement {
         };
       })
     );
-    // TODO (justinfagnani): compile project here
+    this._compileProject();
     this._currentFileIndex = 0;
     // TODO(justinfagnani): whyyyy?
```
e111077, line 289: we can maybe chat about this later?

src/typescript-worker/typescript-worker.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+    return {
+      type: 'bare' as const,
+      url: `https://unpkg.com/${specifier}`,
+    };
+  }
+};
+
+class WorkerLanguageServiceHost implements ts.LanguageServiceHost {
+  compilerOptions: ts.CompilerOptions;
```
e111077, line 300: readonly on these?

src/typescript-worker/typescript-worker.ts
```diff
@@ -0,0 +1,352 @@
...
+  target: ts.ScriptTarget.ES2017,
+  module: ts.ModuleKind.ESNext,
+  experimentalDecorators: true,
+  skipDefaultLibCheck: true,
+  skipLibCheck: true,
+  moduleResolution: ts.ModuleResolutionKind.NodeJs,
+};
+
+/**
+ * Rewrites bare module specifiers to unpkg.com URLs. For now, uses unpkg.com
```
e111077, line 29: unpkg module resolution fails when you point to something like

```
import '@material/mwc-checkbox';
import '@material/mwc-list/mwc-list-item.js`;
```

in this case the paths are weird in such that pointing to a specific file causes a duplication of the `@material/mwc-ripple` file.

src/lib/code-sample-editor.ts
justinfagnani: I don't think there's a security problem with that, and we might need to let the worker URL be specified in some other way at some point, especially if Webpack's `import.meta.url` support is still poor. Aiming to get this working with our sites to test this out.

src/lib/code-sample-editor.ts
justinfagnani: any time :)

src/lib/code-sample-editor.ts
```diff
@@ -284,11 +294,20 @@ export class CodeSampleEditor extends LitElement {
 
   private async _startWorkders() {
```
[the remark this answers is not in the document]
justinfagnani: fixed

src/typescript-worker/typescript-worker.ts
```diff
@@ -0,0 +1,352 @@
+/**
+ * @license
+ * Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
```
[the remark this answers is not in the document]
justinfagnani: fixed

src/typescript-worker/typescript-worker.ts
```diff
@@ -0,0 +1,352 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {TypeScriptWorkerAPI, SampleFile} from '../shared/worker-api.js';
+import {expose} from 'comlink';
+import * as ts from 'typescript';
+
+const compilerOptions = {
```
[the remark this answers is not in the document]
justinfagnani: It's only 6 lines though, I don't think that would help readability.

src/typescript-worker/typescript-worker.ts
```diff
@@ -0,0 +1,352 @@
...
+/**
+ * Rewrites bare module specifiers to unpkg.com URLs. For now, uses unpkg.com
+ * module resolution from there on. We might want to change this and do all
+ * rewrites ourselves to ensure less module duplication.
+ */
+const makeBareSpecifierTransformVisitor = (
+  context: ts.TransformationContext
+): ts.Visitor => {
+  const visitor: ts.Visitor = (node) => {
+    if (ts.isImportDeclaration(node)) {
```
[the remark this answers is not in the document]
justinfagnani: No. I'll file an issue to support those

src/typescript-worker/typescript-worker.ts
```diff
@@ -0,0 +1,352 @@
...
+ * dependencies of the source files, and relies on unpkg.com for transitive
+ * dependencies to load in the preview iframe. In order to enable propert type
+ * checking we will load the entire module graph here.
+ */
+const loadFiles = (
+  files: Array<SampleFile>
+): Promise<Map<string, FileRecord>> => {
+  return new Promise(async (resolve, _reject) => {
+    const fileRecords = new Map<string, FileRecord>();
+    let pendingFileCount = 0;
```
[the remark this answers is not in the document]
justinfagnani: done

src/typescript-worker/typescript-worker.ts
```diff
@@ -0,0 +1,352 @@
...
+            pendingFileCount++;
+
+            // Synchronously set the file to pending so subsequent imports don't
+            // trigger another loading task.
+            fileRecords.set(url, {
+              status: 'pending',
+            });
+
+            // Async task to load the imported file
+            (async () => {
```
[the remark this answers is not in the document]
justinfagnani: Done. It was easier to keep `pendingFileCount` and the call to `resolve()` in the for loop.

src/typescript-worker/typescript-worker.ts
justinfagnani: done

justinfagnani:
Thanks for the review @e111077 PTAL

e111077 (approved):

e111077:
great pass

PR #28, opened 2020-09-11
Support inline files in <script> tags

justinfagnani:
This will make small samples easier to write in markdown files.

cc @arthurevans

### DIFF
--- a/src/lib/code-sample-editor.ts
+++ b/src/lib/code-sample-editor.ts
@@ -94,6 +94,24 @@ const typescriptWorkerScriptUrl = new URL(
  *   }
  * }
  * ```
+ *
+ * Files can also be given as <script> tag children of <code-sample-editor>.
+ * The type attribute must start with "sample/" and then the type of the file,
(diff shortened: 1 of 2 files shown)


src/lib/code-sample-editor.ts
```diff
@@ -434,3 +485,22 @@ const mimeTypeToTypeEnum = (mimeType?: string) => {
...
+
+const typeEnumToMimeType = (type?: string) => {
+  // TODO: infer type based on extension too
+  if (type === undefined) {
+    return;
+  }
+  switch (type) {
+    // TypeScript
+    case 'ts':
+      return 'video/mp2t';
```
aomarks, line 497: Video/mp2t doesn't seem right?

justinfagnani: See the comment above. It's the registered mime-type for .ts files and what I'm seeing returned by servers. afaict there's no mime-type for TypeScript.

aomarks (approved):

PR #37, opened 2020-09-16
Templating / Replacements

e111077:
See readme and demo basically:

`project.json`:

```json
{
  "files": {
    "index.html": {
      "replacements": {
        "<!-- dom body -->": "index-dom.html",
        "<!-- dom head -->": "index-head.html"
      }
    },
    "my-element.js": {},
    "my-second-element.js": {
      "replacements": {
        "/* second element render */": "second-element-render.js"
      }
    }
  }
}

```

`index.html` and `my-second-element.js` are considered templates and the keys will be `string.replace`d sequentially and the templates will not be shown to the user but instead the replacement files.

### TODOS

* allow more metadata for replacements
  * display different title
  * give content type for editor but not fetch (fetching `.html$` will insert `<html><body><head>` into editor)
* Non-sequential replacements
* `replaceAll` instead of `replace` once
* no `await` inside `for` loops (required significant refactoring into ugly code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,70 @@ Or define your project in a JSON manifest:
 }
 ```
 
+### Templating & File Replacements
+
+the JSON manifest supports a simple templating system where you must define an
+insertion point (that is simply fed into `string.replace`) as well as a file
+that will be inserted into that template. Templates will not be displayed to the
+user for editing. For example:
+
+`project.json`:
+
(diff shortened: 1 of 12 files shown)


justinfagnani:
We should still do something like this in the future, but the main branch has diverged too far for this PR. Thanks Elliott!

PR #41, opened 2020-10-02
Allow injected CodeMirror styles to survive lit rendering

aomarks:
Styles were wonky in Firefox and Safari because `EditorView` writes a `<style>` directly into the given root on construction (unless adopted stylesheets are available in which case it uses that, which is why Chrome was not affected). But then lit was rendering and blowing it away.

So my really hacky solution is to snatch the styles `EditorView` injects before the lit render clobber happens, and then tell lit to render them back again itself.

cc @e111077

### DIFF
--- a/src/lib/codemirror-editor.ts
+++ b/src/lib/codemirror-editor.ts
@@ -67,7 +67,7 @@ export class CodeMirrorEditorElement extends LitElement {
   // that we can set this value internally without triggering an update.
   private _value?: string;
 
-  private _capturedCodeMirrorStyles: HTMLStyleElement[] = [];
+  private _capturedCodeMirrorStyles?: NodeListOf<HTMLStyleElement>;
 
   get value() {
     return this._value;
@@ -87,10 +87,7 @@ export class CodeMirrorEditorElement extends LitElement {
   type: 'js' | 'ts' | 'html' | 'css' | undefined;
 
   render() {
-    return html`
-      ${this._capturedCodeMirrorStyles}
-      ${this._editorView?.dom}
-    `;
+    return html` ${this._editorView?.dom} ${this._capturedCodeMirrorStyles} `;
   }
 
   update(changedProperties: PropertyValues) {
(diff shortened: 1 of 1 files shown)


src/lib/codemirror-editor.ts
```diff
@@ -67,6 +67,8 @@ export class CodeMirrorEditorElement extends LitElement {
   // that we can set this value internally without triggering an update.
   private _value?: string;
 
+  private _capturedCodeMirrorStyles: HTMLStyleElement[] = [];
```
justinfagnani, line 70: initialize to `undefined` to skip the lit-html array handling when adoptedStyleSheets is available

src/lib/codemirror-editor.ts
```diff
@@ -127,6 +132,12 @@ export class CodeMirrorEditorElement extends LitElement {
       }),
       root: this.shadowRoot!,
     });
+    // EditorView writes a <style> directly into the given root on construction
+    // (unless adopted stylesheets are available in which case it uses that).
+    // But then lit renders and blows it away. So, we'll just snatch any new
+    // styles beefore this can happen, and then have lit put them back again.
+    this._capturedCodeMirrorStyles = Array.from(
+      this.shadowRoot!.querySelectorAll('style'));
```
justinfagnani, line 140: You don't need the `Array.from()`. querySelectorAll returns a non-live NodeList which is iterable in all the browsers we support (we don't support IE).

justinfagnani, line 140: Do we know that `new EditorView` reuses existing styles so we don't build up an every growing list of styles here every time we create a view?

If we don't, it's ok to leave a TODO to check this for now.

justinfagnani (approved):

justinfagnani:
Seems like a fine solution for now. This won't be an issue with lit-html 2.0

kevinpschaaf:
Should we go ahead and have this rely on lit-html 2.0? Seems like it wouldn't hurt?

src/lib/codemirror-editor.ts
aomarks: Confirmed it re-uses the same style and tested nothing weird happens. Added a comment.

src/lib/codemirror-editor.ts
aomarks: Done.

src/lib/codemirror-editor.ts
aomarks: Done.

aomarks:
> Should we go ahead and have this rely on lit-html 2.0? Seems like it wouldn't hurt?

Hm yeah maybe, though tbh right now I'd rather just get this working on a stable base without any risk of hitting issues related to cutting edge lit

PR #42, opened 2020-10-02
Workaround Safari crash, use MWC canary, build SW, use abs paths in demo

aomarks:
- Add hacky workaround for Safari bug (https://bugs.webkit.org/show_bug.cgi?id=215732) to prevent the browser tab from hard-crashing whenever an `<mwc-tab>` is clicked to change files. (Workaround suggested in https://github.com/material-components/material-components-web-components/issues/1720).

- Update to the latest MWC canary versions to pick up https://github.com/material-components/material-components-web-components/issues/1867, which partly fixes Safari.

- Actually build the service worker, since that wasn't happening before.

- Use absolute paths for the demo project files, since otherwise going to `localhost:8080/demo` will error confusingly while `localhost:8080/demo/` works.

Fixes https://github.com/PolymerLabs/code-sample-editor/issues/38

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,12 +6,13 @@
   "module": "lib/code-sample-editor.js",
   "main": "lib/code-sample-editor.js",
   "scripts": {
-    "build": "npm run build:lib && npm run bundle",
+    "build": "npm run build:lib && npm run build:sw && npm run bundle",
     "build:lib": "tsc --build",
+    "build:sw": "tsc --build src/service-worker",
     "bundle": "rollup -c rollup.config.js",
-    "watch": "tsc --build --watch & rollup -c rollup.config.js -w",
+    "watch": "npm run build:lib -- --watch & npm run build:sw -- --watch & rollup -c rollup.config.js -w",
     "serve": "es-dev-server --node-resolve --event-stream=false",
-    "dev": "npm run watch & npm run serve",
+    "dev": "npm run watch & npm run serve -- --open=demo/",
     "format": "prettier src/**/*.ts --write",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -29,11 +30,11 @@
   },
   "dependencies": {
     "@codemirror/next": "^0.12.0",
-    "@material/mwc-button": "^0.18.0",
-    "@material/mwc-icon-button": "^0.18.0",
-    "@material/mwc-tab": "^0.18.0",
-    "@material/mwc-tab-bar": "^0.18.0",
-    "@material/mwc-textfield": "^0.18.0",
+    "@material/mwc-button": "^0.19.0-canary.615d861d.0",
```
justinfagnani, line 33: with the hashes these might as well be exact versions

src/lib/code-sample-editor.ts
```diff
@@ -50,6 +50,20 @@ declare global {
...
+// to delegatesFocus and shadow roots.
+//
+// https://bugs.webkit.org/show_bug.cgi?id=215732
+// https://github.com/material-components/material-components-web-components/issues/1720
+import {Tab} from '@material/mwc-tab';
+((Tab.prototype as unknown) as {
+  createRenderRoot: Tab['createRenderRoot'];
+  attachShadow: Tab['attachShadow'];
+}).createRenderRoot = function () {
+  return this.attachShadow({mode: 'open', delegatesFocus: false});
```
justinfagnani, line 64: wild 👍

justinfagnani (approved):

package.json

PR #47, opened 2020-10-08
Add tests, fix CSS highlighting

aomarks:
- Sets up @web/test-runner in a GitHub action
- Integration tests for `<code-sample>`
- Unit tests for `<codemirror-editor>`
- Found and fixed a bug that meant CSS files didn't get highlighted

Fixes https://github.com/PolymerLabs/code-sample-editor/issues/35


src/test/code-sample.test.ts
```diff
@@ -0,0 +1,169 @@
+/**
```
justinfagnani, line 1: For consistency, all of our other repos use a `_test.ts` suffix

src/test/code-sample.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+import {assert} from '@esm-bundle/chai';
+import '../lib/code-sample.js';
+import {CodeSampleElement} from '../lib/code-sample.js';
+import {html, render} from 'lit-html';
+import {UpdatingElement} from 'lit-element';
+import {CodeMirrorEditorElement} from '../lib/codemirror-editor.js';
+import {CodeSampleProjectElement} from '../lib/code-sample-project.js';
+
+describe('code-sample', () => {
```
justinfagnani, line 23: Most other places we're using ttd-style test/suite

src/test/code-sample.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  afterEach(() => {
+    container.remove();
+  });
+
+  it('is registered', () => {
+    assert.instanceOf(document.createElement('code-sample'), CodeSampleElement);
+  });
+
+  const pierce = async (...selectors: string[]) => {
```
justinfagnani, line 39: Nice utility. Does open-wc have something like this?

src/test/code-sample.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    });
+    // TODO(aomarks) Chromium and Webkit both fire iframe "load" after the
+    // contentDocument has actually loaded, but Firefox fires it before. Why is
+    // that? If not for that, we wouldn't need to poll here.
+    await new Promise((resolve) => {
+      const check = () => {
+        if (iframe.contentDocument?.body?.textContent?.includes(text)) {
+          resolve();
+        } else {
+          setTimeout(check, 10);
```
justinfagnani, line 73: should you limit the number of times this will retry?

src/test/codemirror-sample.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      for (const child of Array.from(node.childNodes)) {
+        const r = findNodeWithText(child, text);
+        if (r) {
+          return r;
+        }
+      }
+      return undefined;
+    };
+
+    const assertHighlight = async (
```
justinfagnani, line 90: Nice basic test here. If this goes any further I'd reach for screenshot tests. I think there's an issue open on WTR for those.

justinfagnani (approved):

src/test/code-sample.test.ts
aomarks: Done.

src/test/code-sample.test.ts
aomarks: Done.

src/test/code-sample.test.ts
aomarks: I don't see one in https://github.com/open-wc/open-wc/tree/master/packages/testing-helpers

Maybe we should open-source and contribute kevin's shadow-piercer to open-wc or elsewhere.

src/test/code-sample.test.ts
aomarks: Mocha has its own test timeout, so I don't think we need another one. But it does seem like a good idea to kill the polling if the mocha timeout has been reached. Added a variable to stop calling `setTimeout` if `teardown` has been called.

src/test/codemirror-sample.test.ts
aomarks: Agreed.

PR #48, opened 2020-10-08
Support layouts with a dedicated editor for each file

aomarks:
- Add `filename` property/attribute to `<code-sample-editor>` which allows  getting and setting the currently selected file.

- Add `noFilePicker` property (`no-file-picker` attribute) to  `<code-sample-editor>` which disables the top file selection `tab-bar`.

- Add `lineNumbers` property (`line-numbers` attribute) to  `<code-sample-editor>` and `<codemirror-editor>` which enables the left-hand-side gutter with line numbers. Off by default, but switched on in `<code-sample>`.

- The caret is now only displayed when an editor is focused (previously it was always displayed).

So for example, now you can do something like this:

```html
<code-sample-project id="custom-project">
  <script type="sample/html" filename="index.html">
    <!doctype html>
    <body>
      <script src="./index.js">&lt;/script>
    </body>
  </script>

  <script type="sample/ts" filename="index.ts">
    document.body.textContent = 'Hello!';
  </script>
</code-sample-project>

<h4>index.html</h4>
<code-sample-editor
    project="custom-project"
    filename="index.html"
    no-file-picker>
</code-sample-editor>

<h4>index.ts</h4>
<code-sample-editor
    project="custom-project"
    filename="index.ts"
    no-file-picker>
</code-sample-editor>

<h4>preview</h4>
<code-sample-preview
    project="custom-project">
</code-sample-preview>
```

Which will display like this:

image

Fixes https://github.com/PolymerLabs/code-sample-editor/issues/30

### DIFF
--- a/src/lib/code-sample-editor.ts
+++ b/src/lib/code-sample-editor.ts
@@ -85,7 +85,8 @@ export class CodeSampleEditor extends LitElement {
   files?: SampleFile[];
 
   /**
-   * The name of the project file that is currently being displayed.
+   * The name of the project file that is currently being displayed. Set when
+   * changing tabs. Does not reflect to attribute.
    */
   @property()
   filename?: string;
@@ -97,11 +98,11 @@ export class CodeSampleEditor extends LitElement {
   noFilePicker = false;
 
   /**
-   * If true, don't display the left-hand-side gutter with line numbers. Default
-   * false (visible).
+   * If true, display a left-hand-side gutter with line numbers. Default false
+   * (hidden).
    */
-  @property({type: Boolean, attribute: 'no-line-numbers'})
-  noLineNumbers = false;
+  @property({type: Boolean, attribute: 'line-numbers'})
+  lineNumbers = false;
 
   @internalProperty()
   private _currentFileIndex?: number;
@@ -133,16 +134,10 @@ export class CodeSampleEditor extends LitElement {
       this._findProjectAndRegister();
     }
     if (changedProperties.has('files') || changedProperties.has('filename')) {
-      this._currentFileIndex = 0;
-      if (this.filename && this.files) {
-        for (let i = 0; i < this.files.length; i++) {
-          const file = this.files[i];
-          if (file.name === this.filename) {
-            this._currentFileIndex = i;
-            break;
-          }
-        }
-      }
+      this._currentFileIndex =
+        this.files && this.filename
+          ? this.files.map((f) => f.name).indexOf(this.filename)
+          : 0;
       // TODO(justinfagnani): whyyyy?
       if (this._tabBar) {
         await this._tabBar.updateComplete;
@@ -156,9 +151,9 @@ export class CodeSampleEditor extends LitElement {
   render() {
     return html`
       ${this.noFilePicker
-        ? ''
+        ? nothing
         : html` <mwc-tab-bar
-            .activeIndex=${this._currentFileIndex || 0}
+            .activeIndex=${this._currentFileIndex ?? 0}
             @MDCTabBar:activated=${this._tabActivated}
           >
             ${this.files?.map((file) => {
@@ -175,7 +170,7 @@ export class CodeSampleEditor extends LitElement {
         .type=${this._currentFile
           ? mimeTypeToTypeEnum(this._currentFile.contentType)
           : undefined}
-        .noLineNumbers=${this.noLineNumbers}
+        .lineNumbers=${this.lineNumbers}
         @change=${this._onEdit}
       >
       </codemirror-editor>
@@ -184,9 +179,7 @@ export class CodeSampleEditor extends LitElement {
 
   private _tabActivated(e: CustomEvent<{index: number}>) {
     this._currentFileIndex = e.detail.index;
-    this.filename = this.files
-      ? this.files[this._currentFileIndex].name
-      : undefined;
+    this.filename = this.files?.[this._currentFileIndex].name;
   }
 
   private _findProjectAndRegister() {
--- a/src/lib/code-sample-editor.ts
+++ b/src/lib/code-sample-editor.ts
@@ -18,6 +18,7 @@ import {
   customElement,
   css,
   property,
+  internalProperty,
   query,
   PropertyValues,
 } from 'lit-element';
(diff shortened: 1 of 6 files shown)


src/lib/code-sample-editor.ts
```diff
@@ -113,38 +132,50 @@ export class CodeSampleEditor extends LitElement {
     if (changedProperties.has('project')) {
       this._findProjectAndRegister();
     }
-    if (changedProperties.has('files')) {
+    if (changedProperties.has('files') || changedProperties.has('filename')) {
       this._currentFileIndex = 0;
+      if (this.filename && this.files) {
```
justinfagnani, line 137: I think using `indexOf()` would be nice, I'm not sure if using `.map()` would be nice or not:

```ts
this._currentFileIndex = this.files?.map((f) => f.name).indexOf(this.filename) ?? 0;
```

src/lib/code-sample-editor.ts
```diff
@@ -113,38 +132,50 @@ export class CodeSampleEditor extends LitElement {
...
+          const file = this.files[i];
+          if (file.name === this.filename) {
+            this._currentFileIndex = i;
+            break;
+          }
+        }
+      }
       // TODO(justinfagnani): whyyyy?
       if (this._tabBar) {
         await this._tabBar.updateComplete;
```
justinfagnani, line 148: btw, this await was necessary because of some strange issue in mwc-tab-bar, can you check if it's still necessary? Especially because it's bound in the template. I should have opened an issue on MWC...

src/lib/code-sample-editor.ts
```diff
@@ -113,38 +132,50 @@ export class CodeSampleEditor extends LitElement {
...
-          return html`<mwc-tab label=${label}></mwc-tab>`;
-        })}
-        ${this.enableAddFile
-          ? html`<mwc-icon-button icon="add"></mwc-icon-button>`
-          : nothing}
-      </mwc-tab-bar>
+      ${this.noFilePicker
+        ? ''
+        : html` <mwc-tab-bar
+            .activeIndex=${this._currentFileIndex || 0}
```
justinfagnani, line 161: `this._currentFileIndex ?? 0`

src/lib/code-sample-editor.ts
```diff
@@ -83,8 +84,26 @@ export class CodeSampleEditor extends LitElement {
...
+   */
+  @property({type: Boolean, attribute: 'no-file-picker'})
+  noFilePicker = false;
+
+  /**
+   * If true, don't display the left-hand-side gutter with line numbers. Default
+   * false (visible).
+   */
+  @property({type: Boolean, attribute: 'no-line-numbers'})
+  noLineNumbers = false;
```
justinfagnani, line 104: I think you could make this `lineNumbers` and default to `false`.

src/lib/code-sample-editor.ts
```diff
@@ -153,6 +184,9 @@ export class CodeSampleEditor extends LitElement {
 
   private _tabActivated(e: CustomEvent<{index: number}>) {
     this._currentFileIndex = e.detail.index;
+    this.filename = this.files
+      ? this.files[this._currentFileIndex].name
+      : undefined;
```
justinfagnani, line 189:
```suggestion
    this.filename = this.files?.[this._currentFileIndex].name;
```

src/lib/code-sample-editor.ts
```diff
@@ -83,8 +84,26 @@ export class CodeSampleEditor extends LitElement {
   @property({attribute: false})
   files?: SampleFile[];
 
-  // TODO: make a public property/method to select a file
-  @property({attribute: false})
+  /**
+   * The name of the project file that is currently being displayed.
+   */
+  @property()
+  filename?: string;
```
justinfagnani, line 91: This property looks like it's settable externally and internally, which makes me a little wary (consider this element used in a lit-html template with the property being bound - it could get out of sync with the dirty-checking). Document that it's set when changing tabs, and doesn't reflect to the attribute.

src/lib/codemirror-editor.ts
```diff
@@ -58,6 +58,11 @@ export class CodeMirrorEditorElement extends LitElement {
     .cm-wrap {
       height: 100%;
     }
+
+    /* Hide the caret when editor is not focused. */
+    .cm-wrap:not(.cm-focused) .cm-cursor {
+      display: none;
```
justinfagnani, line 64: Does CodeMirror have a programmatic way of hiding the cursor? Wondering in case CM is using a timer to make blinking cursors and whether we want to turn off the timers. It could be using CSS too, just unsure.

justinfagnani (approved):

src/lib/code-sample-editor.ts
aomarks: Can't do exactly that because `this.filename` could be undefined, but I did something similar.

src/lib/code-sample-editor.ts
aomarks: It is still required, the tabs don't render otherwise. I don't understand why either, but I'll look into it another time.

src/lib/code-sample-editor.ts
```diff
@@ -113,38 +132,50 @@ export class CodeSampleEditor extends LitElement {
...
-        ${this.files?.map((file) => {
-          const label = file.name.substring(file.name.lastIndexOf('/') + 1);
-          return html`<mwc-tab label=${label}></mwc-tab>`;
-        })}
-        ${this.enableAddFile
-          ? html`<mwc-icon-button icon="add"></mwc-icon-button>`
-          : nothing}
-      </mwc-tab-bar>
+      ${this.noFilePicker
+        ? ''
```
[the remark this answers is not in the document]
aomarks: Done.

src/lib/code-sample-editor.ts
aomarks: Done.

src/lib/code-sample-editor.ts
aomarks: Done.

src/lib/codemirror-editor.ts
aomarks: I looked into this before, couldn't find anything. It uses a CSS animation though.

src/lib/code-sample-editor.ts
aomarks: Done, but made it on in `<code-sample>`.

src/lib/code-sample-editor.ts
aomarks: Done.

PR #52, opened 2020-10-14
Add pre-render slots to code-sample-editor and preview

aomarks:
Add a `<slot>` to `<code-sample-editor>` and `<code-sample-preview>` which are rendered as-is until those elements are ready. For the editor, this is until the file contents has loaded and the editor has rendered. For the preview, this is until the iframe src has been set and the iframe has fired its "load" event at least one time.

Outstanding issues to address after this PR:

- The file picker tab bar isn't rendered at all before upgrade, so there's some jank when it pops in. I think splitting the file picker into its own element could be a solution here, since then the consumer can set a fixed size for it.

- Same issue for the preview title bar/reload button. For this, I was thinking that reloading should maybe mostly happen automatically based on a debounced timer on the editor, instead of requiring a reload button at all. We could also have a floating reload button in the corner, instead of dedicating a full bar to it, because that way the parent sizing won't be affected when it loads in. Not sure about the title, and whether that's actually needed at all?

Partly addresses https://github.com/PolymerLabs/code-sample-editor/issues/33, but a bit different to the suggestion there.


justinfagnani (approved):
nice!

PR #54, opened 2020-10-16
Polish pass

aomarks:
- Add optional display label for files (e.g. "HTML" and "JS" instead of the filename).
- Add linear progress bar indicator for when preview is loading.
- Automatically reload the preview when an editor is changed (with 500ms debounce).
- Make main layout consistently 70%/30%.
- Various other styling tweaks to make layout more stable.
- Use a resize observer to tell CodeMirror when its parent size changes, because it manages its scrollbars with JavaScript.
- Add some CSS Shadow Parts to allow for more style customization.

image


justinfagnani:
I actually like the preview toolbar because it both makes the alignment of the main areas consistent, and gives a place to show the filename. We may very well want to allow transitions in some demos.

If you look at the React editable samples, they have a toolbar with just the static text "RESULT", probably to balance the layout.

We may also want to add a console or errors tab to that side at some point...

CHANGELOG.md
```diff
@@ -66,6 +66,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - The caret is now only displayed when an editor is on focus (previously it was
   always displayed).
 
+- The `<code-sample>` side-by-side layout is now consistently 70%/30% (widths be
```
justinfagnani, line 69:
```suggestion
- The `<code-sample>` side-by-side layout is now consistently 70%/30% (widths can be
```

src/lib/code-sample-editor.ts
```diff
@@ -153,11 +175,14 @@ export class CodeSampleEditor extends LitElement {
       ${this.noFilePicker
         ? nothing
         : html` <mwc-tab-bar
+            part="file-picker"
             .activeIndex=${this._currentFileIndex ?? 0}
             @MDCTabBar:activated=${this._tabActivated}
           >
             ${this.files?.map((file) => {
-              const label = file.name.substring(file.name.lastIndexOf('/') + 1);
+              const label =
```
justinfagnani, line 183: Does `mwc-tab` support a tooltip where we could put the filename when a label is shown?

src/lib/code-sample-editor.ts
```diff
@@ -213,7 +238,8 @@ export class CodeSampleEditor extends LitElement {
     const value = this._editor.value;
     if (this._currentFile) {
       this._currentFile.content = value!;
-      // TODO: send to worker?
+      console.log('DOIT');
```
justinfagnani, line 241: remove

src/lib/code-sample-project.ts
```diff
@@ -324,7 +328,30 @@ export class CodeSampleProjectElement extends LitElement {
...
+  private _clearSaveTimeout() {
+    if (this._saveTimeoutId !== undefined) {
+      clearTimeout(this._saveTimeoutId);
+      this._saveTimeoutId = undefined;
+    }
+  }
+
+  saveDebounced() {
+    this._clearSaveTimeout();
+    console.log('wa');
```
justinfagnani, line 342: remove

src/lib/code-sample-project.ts
```diff
@@ -324,7 +328,30 @@ export class CodeSampleProjectElement extends LitElement {
...
+    // TODO(aomarks) Consider exposing a property for auto-save timeout, but it
+    // should probably be on the editor or the preview, not the project.
+    this._saveTimeoutId = setTimeout(() => {
+      console.log('wat');
```
justinfagnani, line 346: remove

justinfagnani (changes requested):
Looks good, but want to chat about the layout?

src/lib/code-sample-editor.ts
aomarks: Done.

src/lib/code-sample-project.ts
aomarks: Done.

src/lib/code-sample-project.ts
aomarks: Done.

src/lib/code-sample-editor.ts
aomarks: I kind of feel that if you're using labels, you are typically trying to simplify things by removing the technicality of filenames (e.g. the HTML/JS example in the demo). So I'm not sure I would want to see a filename popup when I interact with the tabs -- seems like a distraction. Can discuss more later, though.

aomarks (commented):
> Looks good, but want to chat about the layout?

Discussed in chat. I added back the toolbar, added a linear-progress loading indicator (see new screenshot), removed the spinning animation on the reload button, and added parts for all the tool bar stuff so that the user can just `display:none` any piece they don't want.

src/lib/code-sample-preview.ts
```diff
@@ -128,19 +142,32 @@ export class CodeSamplePreviewElement extends LitElement {
...
       </div>
-      ${this._iframeLoaded ? nothing : html`<slot></slot>`}
-      <iframe
-        src=${ifDefined(this.src)}
-        @load=${this._onIframeLoad}
-        ?hidden=${!this._iframeLoaded}
-      ></iframe>
+
+      <div id="content">
+        <mwc-linear-progress
```
justinfagnani, line 157: can you put this inside the toolbar instead, and position it absolute so it doesn't reserve space and shift the centering of the other items?

justinfagnani: also, can you make the background transparent so it doesn't show when not active?

aomarks:
> I actually like the preview toolbar because it both makes the alignment of the main areas consistent, and gives a place to show the filename. We may very well want to allow transitions in some demos.

Ok, added it back.

> 
> If you look at the React editable samples, they have a toolbar with just the static text "RESULT", probably to balance the layout.

I made the default text "result", I think that's nicer than "index.html".

> 
> We may also want to add a console or errors tab to that side at some point...

Makes sense, for another PR though.

justinfagnani:
Do you want to remove or make optional the gutter in this PR? I think it'll look much cleaner

src/lib/code-sample-preview.ts
justinfagnani: one more thing: I think it'll look better if the height of the progress bar is the same as the current tab line in the tab bar. I think it's 2px, but hopefully controlled by a css variable.

aomarks: I tried both. I think it looks better below the toolbar, because otherwise 1) it makes the toolbar contents look off-center when it's visible, and 2) this way the user can independently hide the toolbar and progress bar, since they have separate parts. It's already positioned absolutely above the contents of the iframe, and it's already transparent when disabled (that's just the default behavior or linear progress bar, it seems).

aomarks:
> Do you want to remove or make optional the gutter in this PR? I think it'll look much cleaner

It's already optional if you do a cusotm layout

> Do you want to remove or make optional the gutter in this PR? I think it'll look much cleaner

Done.

src/lib/code-sample-preview.ts
justinfagnani: It looks off to me that it doesn't line up with the current tab underline. I think (1) would be alleviated by the smaller height and if the background was transparent (not gray) even when active.

One way to keep it out of the toolbar would be to position it as `top: -2px` (or whatever the hight of the tab bar underline is).

aomarks: Ok, you convinced me, it looks much better in the toolbar once I shrink it to match the height of the tar-bar indicator. See new screenshot. The only downside here is that if you want to disable the toolbar, you have to also tweak the position of the loading indicator, but it's pretty straightforward:

```css
code-sample::part(preview-toolbar) {
  display: none;
}
code-sample::part(preview-loading-indicator) {
  top: 0
}
```

justinfagnani: Thanks for trying it all out!

justinfagnani (approved):

PR #56, opened 2020-10-20
Add configurator UI for github.io

aomarks:
Automatically deploys to https://polymerlabs.github.io/code-sample-editor/

In retrospect, this basically evolved into a primitive knobs system that's really similar to what storybook does. I would be happy to migrate to an existing storybook framework as a followup, preferably an open-sourced version of @rictic 's internal one. I do like the minimalist presentation and the copy/pasteable HTML/CSS panel here, but those both of those seem like things that could probably be replicated in another system.

Customized example: https://polymerlabs.github.io/code-sample-editor/?theme=monokai&fontFamily=Source+Code+Pro&lineNumbers=y&filePickerBackground=1a1a1a&filePickerForeground=f5f5f5&previewToolbarBackground=dedede&previewToolbarForeground=2b2b2b&borders=n&highlight=f0006c&radius=6&barHeight=46&pageBackground=8c8c8c

image

### DIFF
--- a/configurator/project/index.html
+++ b/configurator/project/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html>
+  <head>
+    <script type="module" src="./my-element.js"></script>
+  </head>
+  <body>
+    <div>Here are the contents:</div>
+    <my-element myNumber="5">
+      <div>Element is not upgraded</div>
+    </my-element>
+  </body>
+</html>
--- a/configurator/project/index.html
+++ b/configurator/project/index.html
@@ -4,9 +4,6 @@
     <script type="module" src="./my-element.js"></script>
   </head>
   <body>
-    <div>Here are the contents:</div>
-    <my-element myNumber="5">
-      <div>Element is not upgraded</div>
-    </my-element>
+    <my-element greet="World"></my-element>
   </body>
 </html>
--- a/configurator/project/my-element.ts
+++ b/configurator/project/my-element.ts
@@ -2,14 +2,10 @@ import {LitElement, html, property, customElement} from 'lit-element';
 
 @customElement('my-element')
 export class MyElement extends LitElement {
-  @property({type: Number})
-  myNumber?: number;
+  @property()
+  greet = 'nobody';
 
   render() {
-    return html`
-      <div>Here is ${this.myNumber}</div>
-      <div>And here is my second element:</div>
-      <my-second-element></my-second-element>
-    `;
+    return html`<p>Hello <b>${this.greet}</b>!</p>`;
   }
 }
--- a/configurator/project/my-element.ts
+++ b/configurator/project/my-element.ts
@@ -1,0 +1,15 @@
+import {LitElement, html, property, customElement} from 'lit-element';
+
+@customElement('my-element')
+export class MyElement extends LitElement {
+  @property({type: Number})
+  myNumber?: number;
+
+  render() {
+    return html`
+      <div>Here is ${this.myNumber}</div>
+      <div>And here is my second element:</div>
+      <my-second-element></my-second-element>
+    `;
+  }
+}
--- a/src/configurator/lib/knobs.ts
+++ b/src/configurator/lib/knobs.ts
@@ -1,0 +1,245 @@
+/**
+ * @license
+ * Copyright (c) 2020 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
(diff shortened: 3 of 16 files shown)


configurator/project/my-element.ts
```diff
@@ -0,0 +1,15 @@
...
+@customElement('my-element')
+export class MyElement extends LitElement {
+  @property({type: Number})
+  myNumber?: number;
+
+  render() {
+    return html`
+      <div>Here is ${this.myNumber}</div>
+      <div>And here is my second element:</div>
+      <my-second-element></my-second-element>
```
justinfagnani, line 12: I'd remove this if it isn't included in the project

configurator/project/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html>
+  <head>
+    <script type="module" src="./my-element.js"></script>
+  </head>
+  <body>
+    <div>Here are the contents:</div>
+    <my-element myNumber="5">
+      <div>Element is not upgraded</div>
```
justinfagnani, line 9: I'd suggest we remove this and simplify the example. It was included originally to show the full lifecycle of an element, but I think obscured the demo now.

src/configurator/lib/knobs.ts
```diff
@@ -0,0 +1,245 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+ /**
+  * This kinda turned into something like storybook.js and similar. Maybe we
+  * should just use another library.
```
justinfagnani, line 17: What about Peter's story library?

src/configurator/lib/playground-configurator.ts
```diff
@@ -0,0 +1,478 @@
...
+        return [label, this.sliderKnob(knob)];
+      case 'color':
+        return [label, this.colorKnob(knob)];
+      case 'checkbox':
+        return [label, this.checkboxKnob(knob)];
+    }
+    return '';
+  }
+
+  private selectKnob(knob: KnobsOfType<'select'>) {
```
justinfagnani, line 327: Seems like these should be components... but I'm hesitant to suggest any investment before considering whether to use another library.

One reason _to_ develop a knob system here is to make hybrid source/inspector playgrounds where a sample can define its own knobs and relate them to source, so source and preview can change with the knob values. This could be used to build systems like the previews on material.io: https://material.io/components/buttons

src/configurator/lib/playground-configurator.ts
```diff
@@ -0,0 +1,478 @@
...
+        }}
+      />
+    `;
+  }
+
+  /**
+   * https://tholman.com/github-corners/
+   */
+  private get githubCorner() {
+    return html`<a
```
justinfagnani, line 417: This doesn't need to be a method, and it might be nice to put it in another file... or could stay just like this.

justinfagnani (approved):

justinfagnani:
This is really, really awesome @aomarks !

src/configurator/lib/knobs.ts
aomarks: Yeah, I mentioned that in the PR description. If we opened sourced that, I would prefer to use it.

configurator/project/index.html
aomarks: Done, simplified to a "Hello World" example.

configurator/project/my-element.ts
aomarks: Done.

src/configurator/lib/playground-configurator.ts
aomarks: Moved to a const.

src/configurator/lib/playground-configurator.ts
aomarks: > Seems like these should be components... but I'm hesitant to suggest any investment before considering whether to use another library.

Yeah, I'd like to use Peter's knob system if it was open source. Agree I don't want to invest more in this, it works fine for what we need right now.

> One reason _to_ develop a knob system here is to make hybrid source/inspector playgrounds where a sample can define its own knobs and relate them to source, so source and preview can change with the knob values. This could be used to build systems like the previews on material.io: https://material.io/components/buttons

Interesting idea, will think about it.

PR #57, opened 2020-10-21
Rename to playground-elements

aomarks:
Rename elements:

- `<playground-ide>`
  - I'm not 100% on this, but the reasoning is that it's kinda memorable, and it's the one that integrates the other elements.
  - Other options: `<playground-sample>`, `<playground-layout>`, `<playground-side-by-side>`, ...
- `<playground-project>`
- `<playground-file-editor>`
- `<playground-code-editor>`
  - This used to be the codemirror element. Maintaining this pure code editor which doesn't care about project files, is still a nice factoring I think. I hope the `file` vs `code` distinction works here.
- `<playground-preview>`

Also renames the package to `playground-elements`, and prepares the `v0.1.0` release. Once this PR is approved, I'll rename the Git repo to `PolymerLabs/playground-elements`.


justinfagnani:
Some thoughts on the "main" element name:
- `<playground-project>` and rename project to `<playground-loader>`
- `<playground-app>`?
- `<playground-demo>`

For `<playground-file-editor>`, I would consider naming that something that indicates there can be more than one file. Maybe `<playground-editor-group>`? And then there could be `<playground-editor>` for a single file. I know the factoring doesn't quite support that naming now, but maybe things can be factored so that there's one single-file editor and another component that adds the tab bar.

aomarks:
> Some thoughts on the "main" element name:
> 
> * `<playground-project>` and rename project to `<playground-loader>`
> * `<playground-app>`?
> * `<playground-demo>`

Not bad, but they maybe sound a little broad. If we add a `<playground-tutorial>` or similar, then there would be two self-contained top-level elements, and it seems helpful for the names to distinguish between those. So that's why I was thinking `ide`, or `side-by-side`, or something that kind of describes the particular use-case/layout.

> 
> For `<playground-file-editor>`, I would consider naming that something that indicates there can be more than one file. Maybe `<playground-editor-group>`? And then there could be `<playground-editor>` for a single file. I know the factoring doesn't quite support that naming now, but maybe things can be factored so that there's one single-file editor and another component that adds the tab bar.

I was actually thinking of fully factoring out the file picker to `<playground-file-picker>`, both to allow more layout flexibility, and because you could imagine dropping in a different kind of file picker (e.g. a dropdown menu, or a vertical tab bar).

Also, I do quite like the `<playground-file-editor>` vs `<playground-code-editor>` distinction, where the latter is a pure codemirror element with a `value` property, and the former adds synchronization with project files. For example, I use `<playground-code-editor readonly>` in the configurator to simply syntax-highlight some HTML/CSS -- I don't really want to require a project for that.

Here's how I was thinking about each piece as a diagram:

image

justinfagnani:
That seems pretty nice. One small suggestion would be to separate out a `<playground-preview-toolbar>` similar to the `<playground-file-picker>` so that you could use a preview w/o a toolbar, swap in a different toolbar, and layout the 4 components of the IDE with CSS grid.

I'd also consider naming the current file picker something related to this specific implementation if you're thinking of other options already, maybe like `<playground-file-tabs>`?

README.md
```diff
@@ -1,19 +1,24 @@
-# code-sample-editor
+# playground-elements
+
+[![Published on npm](https://img.shields.io/npm/v/code-sample-editor.svg)](https://www.npmjs.com/package/code-sample-editor) ![Tests](https://github.com/PolymerLabs/code-sample-editor/workflows/Tests/badge.svg)
```
justinfagnani, line 3: Update `code-sample-editor` package name here

README.md
```diff
@@ -93,30 +104,34 @@ Or define your project in a JSON manifest:
...
 ```html
-<code-sample-editor project-src="./example-1.json"></code-sample-editor>
+<playground-ide project-src="./example-1.json"></playground-ide>
 ```
 
-`<code-sample-editor>` uses bare module specifiers in its code, so you'll need a server that supports rewriting module specifiers with the Node module resolution algorithm, or a build tool like Rollup.
+`<playground-ide>` uses bare module specifiers in its code, so you'll need a server that supports rewriting module specifiers with the Node module resolution algorithm, or a build tool like Rollup.
 
-`<code-sample-editor>` also uses `import.meta.url` to load the worker scripts. note that Webpack does not support that currently.
+`<playground-ide>` also uses `import.meta.url` to load the worker scripts. note that Webpack does not support that currently.
```
justinfagnani, line 129: WebPack 5 does support `import.meta.url` now

demo/index.html
```diff
@@ -4,12 +4,12 @@
 <head>
   <link href="https://fonts.googleapis.com/css?family=Material+Icons&display=block" rel="stylesheet">
   <script src="https://unpkg.com/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
-  <script type="module" src="../lib/code-sample.js"></script>
+  <script type="module" src="../lib/playground-ide.js"></script>
```
justinfagnani, line 7: I think we should lift the public entrypoints out of the lib/ folder

package.json
```diff
@@ -1,10 +1,12 @@
...
+  "name": "playground-elements",
+  "version": "0.1.0",
   "description": "A multi-file code editor component with live preview",
+  "homepage": "https://github.com/PolymerLabs/playground-elements#readme",
+  "repository": "github:PolymerLabs/playground-elements",
   "type": "module",
-  "module": "lib/code-sample-editor.js",
-  "main": "lib/code-sample-editor.js",
+  "module": "lib/playground-ide.js",
+  "main": "lib/playground-ide.js",
```
justinfagnani, line 9: might want to make an index.js file so -ide doesn't always have the be the aggregating module.

src/lib/playground-file-editor.ts
```diff
@@ -26,9 +26,9 @@ import '@material/mwc-tab-bar';
 import {TabBar} from '@material/mwc-tab-bar';
 import '@material/mwc-tab';
 import {SampleFile} from '../shared/worker-api.js';
-import {CodeSampleProjectElement} from './code-sample-project';
-import './codemirror-editor.js';
-import {CodeMirrorEditorElement} from './codemirror-editor.js';
+import {PlaygroundProject} from './playground-project';
```
justinfagnani, line 29: add `.js` extension to import

src/lib/playground-ide.ts
```diff
@@ -14,17 +14,17 @@
...
+import './playground-project.js';
+import './playground-file-editor.js';
+import './playground-preview.js';
 
 /**
  * A multi-file code editor component with live preview that works without a
  * server.
  *
- * <code-sample> loads a project configuration file and the set of source files
+ * <playground> loads a project configuration file and the set of source files
```
justinfagnani, line 25: Change to `<playground-ide>` here and 4 places below

justinfagnani (approved):

demo/index.html
aomarks: Done.

package.json

src/lib/playground-ide.ts
aomarks: Oops, done.

aomarks (commented):
> That seems pretty nice. One small suggestion would be to separate out a `<playground-preview-toolbar>` similar to the `<playground-file-picker>` so that you could use a preview w/o a toolbar, swap in a different toolbar, and layout the 4 components of the IDE with CSS grid.

Agreed. I'm going to do this refactor in a follow up PR.

> 
> I'd also consider naming the current file picker something related to this specific implementation if you're thinking of other options already, maybe like `<playground-file-tabs>`?

Also agreed.

src/lib/playground-file-editor.ts
aomarks: Done.

PR #61, opened 2020-10-25
Allow resizing

aomarks:
- Added `--playground-preview-size` CSS custom property to control the initial size of the RHS preview panel in `<playground-ide>`. The LHS editor will take up all additional space. Defaults to `30%`.

- Added `resizable` property/attribute which allows users to click and drag in the space between the LHS editor and RHS preview of `<playground-ide>` to change their relative sizes.

- **Try it out at https://polymerlabs.github.io/playground-elements/?resizable=y**

Also:
- Fix some invalid module import paths.
- TypeScript decorator runtime is now imported from `tslib` instead of inlined.
- Fix reload button not respecting `--playground-preview-toolbar-foreground-color`.
- Prepare `v0.1.1` release.

image

Fixes https://github.com/PolymerLabs/playground-elements/issues/60
Fixes https://github.com/PolymerLabs/playground-elements/issues/62

### DIFF
--- a/src/playground-ide.ts
+++ b/src/playground-ide.ts
@@ -227,7 +227,7 @@ export class PlaygroundIde extends LitElement {
       // Note we set this property on the RHS element instead of the host so
       // that when "resizable" is toggled, we don't reset a host value that the
       // user might have set.
-      this._rhs.style.removeProperty('--playground-preview-width');
+      this._rhs?.style.removeProperty('--playground-preview-width');
     }
     super.update(changedProperties);
   }
--- a/src/playground-ide.ts
+++ b/src/playground-ide.ts
@@ -232,9 +232,9 @@ export class PlaygroundIde extends LitElement {
     super.update(changedProperties);
   }
 
-  private onResizeBarPointerdown(event: PointerEvent) {
+  private onResizeBarPointerdown({pointerId}: PointerEvent) {
     const bar = this._resizeBar;
-    bar.setPointerCapture(event.pointerId);
+    bar.setPointerCapture(pointerId);
 
     const rhsStyle = this._rhs.style;
     const {left: hostLeft, right: hostRight} = this.getBoundingClientRect();
@@ -253,6 +253,7 @@ export class PlaygroundIde extends LitElement {
     bar.addEventListener('pointermove', onPointermove);
 
     const stopDragging = () => {
+      bar.releasePointerCapture(pointerId);
       bar.removeEventListener('pointermove', onPointermove);
       bar.removeEventListener('pointerup', stopDragging);
     };
--- a/src/playground-ide.ts
+++ b/src/playground-ide.ts
@@ -174,8 +174,8 @@ export class PlaygroundIde extends LitElement {
   @property()
   theme = 'default';
 
-  @query('#resizeOverlay')
-  private _resizeOverlay!: HTMLDivElement;
+  @query('#resizeBar')
+  private _resizeBar!: HTMLDivElement;
 
   @query('#rhs')
   private _rhs!: HTMLDivElement;
@@ -204,10 +204,9 @@ export class PlaygroundIde extends LitElement {
       <div id="rhs">
         ${this.resizable
           ? html`<div
-                id="resizeBar"
-                @mousedown=${this.onResizeBarMousedown}
-              ></div>
-              <div id="resizeOverlay"></div>`
+              id="resizeBar"
+              @pointerdown=${this.onResizeBarPointerdown}
+            ></div>`
           : nothing}
 
         <playground-preview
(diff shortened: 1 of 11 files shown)


e111077:
this is amazing like omg

src/playground-ide.ts
```diff
@@ -147,17 +201,67 @@ export class PlaygroundIde extends LitElement {
...
+          location="Result"
+          .project=${projectId}
+        ></playground-preview>
+      </div>
     `;
   }
+
+  async update(changedProperties: PropertyValues<this>) {
+    super.update(changedProperties);
+    if (changedProperties.has('resizable') && this.resizable === false) {
```
e111077, line 228: this is a slightly pedantic suggestion, but generally most work inside `update` IMO should go before `super.update` as to accidentally prevent a rerender cycle. Pedantic because this conditional doesn't actually cause a rerender, but future work that may trigger render cycle, people may just add onto this.

src/playground-ide.ts
```diff
@@ -147,17 +201,67 @@ export class PlaygroundIde extends LitElement {
...
+    const rhsMinWidth = 100;
+    const rhsMaxWidth = hostWidth - 100;
+
+    const onMousemove = (event: MouseEvent) => {
+      const rhsWidth = Math.min(
+        rhsMaxWidth,
+        Math.max(rhsMinWidth, hostRight - event.clientX)
+      );
+      const percent = (rhsWidth / hostWidth) * 100;
+      rhsStyle.setProperty('--playground-preview-width', `${percent}%`);
```
e111077, line 252: why not use a declarative method of `stylemap` + an `@internalProperty`?

src/playground-ide.ts
```diff
@@ -147,17 +201,67 @@ export class PlaygroundIde extends LitElement {
...
+    overlay.addEventListener('mousemove', onMousemove);
+
+    const stopDragging = () => {
+      overlay.classList.remove('resizing');
+      overlay.removeEventListener('mousemove', onMousemove);
+      overlay.removeEventListener('mouseup', stopDragging);
+      overlay.removeEventListener('mouseout', stopDragging);
+    };
+    overlay.addEventListener('mouseup', stopDragging);
+    overlay.addEventListener('mouseout', stopDragging);
```
e111077, line 263: why not use pointerevents with capture to prevent this weird use case? It's now supported on safari (though unfortunately not yet safari -2 versions)

e111077 (changes requested):

src/playground-ide.ts
aomarks: Because I'd never heard of that before. Thanks! Switched, tested in latest Chrome/Firefox/Safari.

src/playground-ide.ts
aomarks: Done.

aomarks:
Also fixed the reload button color issue.

src/playground-ide.ts
```diff
@@ -147,17 +201,63 @@ export class PlaygroundIde extends LitElement {
...
+        Math.max(rhsMinWidth, hostRight - event.clientX)
+      );
+      const percent = (rhsWidth / hostWidth) * 100;
+      rhsStyle.setProperty('--playground-preview-width', `${percent}%`);
+    };
+    bar.addEventListener('pointermove', onPointermove);
+
+    const stopDragging = () => {
+      bar.removeEventListener('pointermove', onPointermove);
+      bar.removeEventListener('pointerup', stopDragging);
```
e111077, line 257: I'm not sure, but should you be [releasing pointer capture](https://developer.mozilla.org/en-US/docs/Web/API/Element/releasePointerCapture)? Not sure if it stays in memory like resize observers after references are GC'd. @justinfagnani any idea?

e111077 (changes requested):

src/playground-ide.ts
aomarks: Seems reasonable, done.

e111077 (approved):

src/playground-ide.ts
```diff
@@ -91,6 +108,30 @@ export class PlaygroundIde extends LitElement {
...
     }
+
+    #resizeBar {
+      position: absolute;
+      top: 0;
+      left: -5px;
+      width: 10px;
+      height: 100%;
+      z-index: 9;
+      cursor: col-resize;
```
justinfagnani, line 119: I think it'd be nice to have a default hover-over highlighting here. I find that it feels better when the cursor doesn't just change on otherwise unindicated regions

src/playground-ide.ts
```diff
@@ -147,17 +201,64 @@ export class PlaygroundIde extends LitElement {
...
+    if (changedProperties.has('resizable') && this.resizable === false) {
+      // Note we set this property on the RHS element instead of the host so
+      // that when "resizable" is toggled, we don't reset a host value that the
+      // user might have set.
+      this._rhs?.style.removeProperty('--playground-preview-width');
+    }
+    super.update(changedProperties);
+  }
+
+  private onResizeBarPointerdown({pointerId}: PointerEvent) {
```
justinfagnani, line 235: I would still name private member with a `_` prefix for the benefit of JS consumers, devtools, etc.

justinfagnani (commented):
A couple of tiny post-merge comments... LGTM!

src/playground-ide.ts
aomarks: Can you clarify? Do you mean that hovering over the resize bar should change the color of something, but not change the cursor? Which element's color should change?

PR #63, opened 2020-10-28
Custom property based theming

aomarks:
Replaces the `theme` property (which only supported two hard-coded themes), with a pure custom property based system with all standard CodeMirror themes included.

**Try it out at https://polymerlabs.github.io/playground-elements/?theme=ayu-mirage**

<hr>

- Adds CSS Custom Properties for configuring the CodeMirror theme.

  Includes broad properties like `--playground-code-background` and
  `--playground-selection-color`, as well as token-specific properties like
  `--playground-code-keyword-color` and `--playground-code-string-color`.

- Adds `themes/` directory which includes CSS files and JS modules for each of
  the standard CodeMirror themes, adapted to use CSS Custom Properties. These
  files are **generated** using a playwright script that tests each stylesheet against
  some mock CodeMirror DOM.

  To load a theme, load its stylesheet into your document or shadow root scope,
  and set the corresponding `playground-theme-NAME` class on any playground
  element or ancestor.

  In HTML:

  ```html
  <head>
    <link
      rel="stylesheet"
      href="/node_modules/playground-elements/themes/mbo.css"
    />
  </head>

  <body class="playground-theme-mbo">
    <playground-ide></playground-ide>
  </body>
  ```

  In a LitElement:

  ```js
  import mbo from 'playground-elements/themes/mbo.css.js';
  class MyElement extends LitElement {
    static styles = [mbo, css`...`];
    render() {
      return `
        <playground-ide class="playground-theme-mbo">
        </playground-ide>
      `;
    }
  }
  ```

Here's an example of a generated theme (26 properties):

```css
.playground-theme-ayu-mirage {
  --playground-code-background: rgb(31, 36, 48);
  --playground-code-default-color: rgb(203, 204, 198);
  --playground-code-cursor-color: rgb(255, 204, 102);
  --playground-code-selection-background: rgb(52, 69, 90);
  --playground-code-gutter-background: rgb(31, 36, 48);
  --playground-code-gutter-border-right: 0px;
  --playground-code-linenumber-color: rgb(61, 66, 77);
  --playground-code-atom-color: rgb(174, 129, 255);
  --playground-code-attribute-color: rgb(255, 213, 128);
  --playground-code-builtin-color: rgb(255, 204, 102);
  --playground-code-comment-color: rgb(92, 103, 115);
  --playground-code-def-color: rgb(255, 213, 128);
  --playground-code-keyword-color: rgb(255, 167, 89);
  --playground-code-meta-color: rgb(203, 204, 198);
  --playground-code-number-color: rgb(255, 204, 102);
  --playground-code-operator-color: rgb(203, 204, 198);
  --playground-code-qualifier-color: rgb(203, 204, 198);
  --playground-code-string-color: rgb(186, 230, 126);
  --playground-code-string-2-color: rgb(203, 204, 198);
  --playground-code-tag-color: rgb(92, 207, 230);
  --playground-code-type-color: rgb(255, 167, 89);
  --playground-code-variable-color: rgb(203, 204, 198);
  --playground-code-variable-2-color: rgb(242, 135, 121);
  --playground-code-variable-3-color: rgb(92, 207, 230);
  --playground-code-callee-color: rgb(203, 204, 198);
  --playground-code-property-color: rgb(242, 158, 116)
}
```


e111077:
spicy!

e111077:
On custom props vs theme: Are we afraid of reaching this supposed css custom prop performance threshold that Dan keeps warning me that youtube reached? In MWC we use custom props because it makes it easier to maintain with the one version policy. If this ends up living only externally, would it be worth weighing this contribution to that perf threshold?

scripts/theme-generator.js
```diff
@@ -36,6 +36,13 @@ import * as webDevServer from '@web/dev-server';
 import * as playwright from 'playwright';
 import CleanCSS from 'clean-css';
 
+// csso
```
justinfagnani, line 39: remove?

aomarks:
> On custom props vs theme: Are we afraid of reaching this supposed css custom prop performance threshold that Dan keeps warning me that youtube reached? In MWC we use custom props because it makes it easier to maintain with the one version policy. If this ends up living only externally, would it be worth weighing this contribution to that perf threshold?

Maybe? I'm a little skeptical that ~30 properties could cause a problem (especially as a proportion of all the other work this component already does), but I guess only a benchmark could really tell us.

I did shrink the number of properties to ~34 to ~26 by finding some that weren't actually being used by our particular highlighters, and collapsing a few selection related ones into a single property. There are a few more token classes that we could potentially collapse (e.g. `meta` which is only used for `<!doctype html>` in HTML, and `qualifier` which is only used for `.class` selectors in CSS).

I added documentation of all these properties in the README, with examples in each language.

justinfagnani (approved):
Awesome!

scripts/theme-generator.js
aomarks: Done.

PR #79, opened 2020-11-11
Support two-origin deployment

aomarks:
- The preview iframe and service worker can now execute on a separate origin to the project, and is encouraged with a console warning.
- The origin is changed by setting a new `sandboxBaseUrl` property.
- It works by using another iframe, in the project element, that coordinates with the service worker.

Also:
- Rename `service-worker.js` and `typescript-worker.js` to `playground-service-worker.js` and `playground-typescript-worker.js`.
- Minify worker scripts.

Fixes https://github.com/PolymerLabs/playground-elements/issues/69
Fixes https://github.com/PolymerLabs/playground-elements/issues/64
Fixes https://github.com/PolymerLabs/playground-elements/issues/73


rollup.config.js
```diff
@@ -15,7 +16,54 @@ export function simpleReplace(replacements) {
   };
 }
 
+function inlineHtml(htmlOutPath) {
```
justinfagnani, line 19: can you add comments here, or below where it's used, on what this is for?

src/playground-ide.ts
```diff
@@ -140,12 +140,32 @@ export class PlaygroundIde extends LitElement {
...
+   *
+   * It is highly advised to change this property to a URL on a separate origin
+   * which has no privileges to perform sensitive actions or access sensitive
+   * data. This is because this element will execute arbitrary JavaScript, and
+   * does not have the ability to sanitize or sandbox it.
+   *
+   * This URL must host the following files from the playground-elements
+   * package:
+   *   1. playground-service-worker.js
+   *   2. playground-service-worker-proxy.html
```
justinfagnani, line 154: Do you think it'd be easier to follow if the attribute pointed directly to the SW-registering HTML file?

src/playground-project.ts
```diff
@@ -147,7 +181,13 @@ export class PlaygroundProject extends LitElement {
   }
 
   render() {
-    return html`<slot @slotchange=${this._slotChange}></slot>`;
+    return html`
+      <slot @slotchange=${this._slotChange}></slot>
+      <iframe
```
justinfagnani, line 186: Can you add a comment somewhere on the purpose of this iframe? Do we have architecture docs anywhere?

src/playground-project.ts
```diff
@@ -223,79 +263,48 @@ export class PlaygroundProject extends LitElement {
...
-  }
-
-  private async _installServiceWorker() {
-    if (!('serviceWorker' in navigator)) {
-      // TODO: show this in the UI
-      console.warn('ServiceWorker support required for <playground-project>');
+  private _onServiceWorkerProxyIframeLoad() {
+    const window = this._iframe.contentWindow;
+    if (!window) {
+      console.error('IFrame did not have window');
```
justinfagnani, line 274: Use `console.assert()` or throw? Can the message be a bit more descriptive? which iframe in particular.

src/playground-project.ts
```diff
@@ -223,79 +263,48 @@ export class PlaygroundProject extends LitElement {
...
-    } else {
-      console.warn('unhandled service worker registration state', registration);
-    }
+    port1.start();
+    const initMessage: ProxyInitMessage = {
+      port: port2,
+      url: 'playground-service-worker.js',
+      scope: this.sandboxScope,
+    };
+    window.postMessage(initMessage, '*', [initMessage.port]);
```
justinfagnani, line 290: We can do better than `'*'` now, right? Does it matter?

justinfagnani (approved):
Looks great. Just a couple of non-blocking questions and it wouldn't be bad to have a few more comments on the initialization workflow, though it's followable as is.

rollup.config.js
aomarks: Done.

src/playground-ide.ts
aomarks: As written, you can just point it at a URL that is hosting `node_modules/playground-elements`, and don't have to know what the name of the files should be. This documentation is here just so people know what needs to be there when they are building/deploying. Seems simpler this way? Not sure why somebody would want to change the HTML file name.

src/playground-project.ts
aomarks: Added brief comment. No docs, but I was going to update the diagram and include it in the README when I get around to that.

src/playground-project.ts
aomarks: Done. I don't think this error will ever actually happen, but can't be sure.

src/playground-project.ts
aomarks: We could constrain it to the configured origin, but that would add the limitation that the URL may not redirect to another origin. I'm not sure what security issue this would protect against, and could be an inconvenience. It seems like if an attacker was able to redirect that URL somewhere else, then all bets are off anyway?

PR #83, opened 2020-12-01
fix(preview): Safari scrolling bug

e111077:
For some reason, in Safari, if the rendered content in the iframe has `height: 100vh` every time you interact with the code on the left, the right side preview continues to grow.


aomarks:
Mind updating changelog?

e111077:
updated!

aomarks (approved):

PR #86, opened 2020-12-03
Fix project fetch race conditions, and handle service worker termination

aomarks:
- Fix race condition that caused preview to 404. Occurred when `project-src` was set, plus any child was slotted _even if only whitespace_, and the project files fetch took longer than the service worker handshake.

- Fix race condition that caused preview to never load. Occured when `project-src` was set, with no slotted child, and the project files fetch took longer than the service worker handshake.

- Handle service worker terminations. Service workers are allowed to terminate anytime but this doesn't count as a state change, so we weren't noticing it before. Any fetch/message event wakes the service worker up again, however it will have lost all of its state, which includes our file APIs. We now detect when this has happened, send a postMessage to request a reconnect, and block fetch responses until we receive the file API.

Also:

- Serve our own 404s instead of forwarding to the server. Since we know we're serving the virtual file system based on the URL, there's no reason to fallback to the server. There's no guarantee the server will return a 404, plus even if it does, it doesn't make sense to display the 404 in the style of the server.

- Refactor post messages to be more strongly typed and easier to follow.

- Strip web-dev-server reload inject script more reliably (it changed recently).

Hopefully fixes https://github.com/PolymerLabs/playground-elements/issues/39
Hopefully fixes https://github.com/PolymerLabs/playground-elements/issues/85

### DIFF
--- a/src/playground-service-worker-proxy.ts
+++ b/src/playground-service-worker-proxy.ts
@@ -12,13 +12,12 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {ESTABLISH_HANDSHAKE} from './shared/worker-api.js';
-
-export interface ProxyInitMessage {
-  port: MessagePort;
-  url: string;
-  scope: string;
-}
+import {
+  PlaygroundMessage,
+  CONNECT_SW_TO_PROJECT,
+  CONNECT_PROJECT_TO_SW,
+  CONFIGURE_PROXY,
+} from './shared/worker-api.js';
 
 (async () => {
   try {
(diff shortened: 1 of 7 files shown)


aomarks:
Well, I can confirm that this doesn't fix the problem (or at least, not all of them), plus this breaks safari and firefox :) Will debug further tomorrow.

aomarks:
> Well, I can confirm that this doesn't fix the problem (or at least, not all of them), plus this breaks safari and firefox :) Will debug further tomorrow.

Found a new race condition, and removed an earlier fix that wasn't useful and broke Safari/Firefox.

src/playground-service-worker-proxy.ts
```diff
@@ -64,6 +78,19 @@ export interface ProxyInitMessage {
     }
   });
 
+  navigator.serviceWorker.addEventListener(
+    'message',
+    (event: MessageEvent<PlaygroundMessage>) => {
+      if (event.data.type === MISSING_FILE_API && registration.active) {
+        // A fetch was made for a session that the service worker doesn't have a
```
e111077, line 85: 👏 👏 👏 👏 👏

e111077 (approved):
So many much-needed edge cases addressed in this PR! Amazing work for one night!

src/service-worker/playground-service-worker.ts
```diff
@@ -13,34 +13,82 @@
...
-const getFile = async (e: FetchEvent, path: string, sessionId: SessionID) => {
-  const fileAPI = fileAPIs.get(sessionId);
+const findSessionProxyClient = async (
+  sessionId: string
+): Promise<Client | undefined> => {
+  for (const client of await self.clients.matchAll({
+    includeUncontrolled: true,
+  })) {
+    if (
+      new URL(client.url).searchParams.get('playground-session-id') ===
```
justinfagnani, line 58: nice

justinfagnani (approved):
+100!

PR #88, opened 2020-12-08
Fix shifted/hidden code in Safari with line numbers

aomarks:
CodeMirror likes to be told when it resizes to adjust some of its styling parameters. We already had a resize observer running for this purpose, but it was incorrectly calling `setSize` instead of `refresh` (https://codemirror.net/doc/manual.html#refresh). This probably affected Safari but not Chrome just because of some timing differences in when layout happens.


e111077 (approved):
You are god mode

PR #89, opened 2021-01-07
Add files property to project and ide

aomarks:
Adds `files` property to `<playground-ide>` and `<playground-project>`.

We can use this property to directly get and set the array of project files (e.g. to save/restore the project state to JSON).

Note that directly setting `files` always takes precedence over setting `projectSrc`, and both take precedence over slotted children.

Also inlined the reload icon SVG so that users don't need to depend on the Material Icons font (since it's inconvenient, and much larger than this one SVG).

Release v0.3.7


src/playground-ide.ts
```diff
@@ -188,17 +190,31 @@ export class PlaygroundIde extends LitElement {
   @property({type: Boolean})
   resizable = false;
 
+  /**
+   * Get or set the array of project files.
+   *
+   * Files set through this property always take precedence over `projectSrc`
+   * and slotted children.
+   */
+  @property({type: Object})
```
justinfagnani, line 199: is `{type: Object}` intentional? If you're not deserializing from JSON it doesn't buy you anything over `attribute: false`, but you might be deserializing from JSON here...

src/playground-preview.ts
```diff
@@ -160,10 +160,21 @@ export class PlaygroundPreview extends LitElement {
...
-        ></mwc-icon-button>
+        >
+          <svg
+            xmlns="http://www.w3.org/2000/svg"
+            viewBox="0 0 24 24"
+            fill="currentcolor"
+            width="18px"
+            height="18px"
+          >
+            <path
```
justinfagnani, line 173: I'd love to see a comment with the a link to the source of the path

src/playground-project.ts
```diff
@@ -160,6 +160,30 @@ export class PlaygroundProject extends LitElement {
...
 
+  private _filesSetExternally = false;
+
+  /**
+   * Get or set the array of project files.
+   *
+   * Files set through this property always take precedence over `projectSrc`
+   * and slotted children.
+   */
+  @property({type: Object})
```
justinfagnani, line 171: if you're not using the attribute, you don't really need the decorator, you could just call `requestUpdate` in the setter.

src/playground-project.ts
```diff
@@ -270,12 +295,18 @@ export class PlaygroundProject extends LitElement {
         };
       })
     );
-    this._compileProject();
+    if (!this._filesSetExternally) {
```
justinfagnani, line 298: should this check come earlier? It looks like if `this._filesSetExternally` is true you fetch the files then throw them away

src/playground-ide.ts
aomarks: Switched to `attribute: false`.

src/playground-preview.ts
aomarks: Added comment.

src/playground-project.ts
aomarks: Done.

src/playground-project.ts
aomarks: Just being defensive in case files was explicitly set after the project fetch started, but before it completed, to make sure the "explicitly set always wins" rule always holds. Added a comment.

aomarks: Oh and, it is also checked before the fetch starts.

justinfagnani: ah, ok

justinfagnani (approved):

aomarks:
Added a hack to fix flex layouts.

PR #90, opened 2021-01-14
Support adding/deleting/renaming files

aomarks:
Demo: https://polymerlabs.github.io/playground-elements/?editableFileSystem=y

image

- Added `editableFileSystem` property (`editable-file-system` attribute) to `<playground-ide>`. When `true`, the user can create, delete, and rename files in the virtual filesystem.

  - To create a new file: click the "+" icon button in the tab bar.
  - To rename a file: click the three-dots menu button on its tab, and select "Rename".
  - To delete a file: click the three-dots menu button on its tab, and select "Delete".

- Added `<playground-file-system-controls>` element which displays popup menus/dialogs for creating, deleting, and renaming files.

- Refactored the way connecting-to-projects work, using a new `PlaygroundConnectedElement` base class.


justinfagnani:
Nice!

Some ux/style suggestions:

1. Increase density / reduce font size of menu
2. Anchor menu to the bottom of the button so the button still shows when the menu is open
3. Only show the menu button on hover over
4. Add icons to menu: baseline_delete_black_18dp and baseline_edit_black_18dp

src/playground-tab-bar.ts
```diff
@@ -0,0 +1,376 @@
...
+              @newFile=${this._onNewFile}
+            >
+            </playground-file-system-controls>
+          `
+        : nothing}
+    `;
+  }
+
+  async updated() {
+    // TODO(aomarks) There still seems to be a timing bug where the mwc-tab-bar
```
justinfagnani, line 190: Do we have an MWC issue for this bug?

src/playground-tab-bar.ts
```diff
@@ -0,0 +1,376 @@
...
+    // The active file was deleted, and another file has now taken the tab
+    // position of the old one. Select that file.
+    this._activeFileName = this._files[this._activeFileIndex].name;
+  }
+}
+
+/**
+ * Internal element for tabs within <playground-tab-bar>.
+ *
+ * This is a subclass of <mwc-tab>. We subclass because <mwc-tab> only supports
```
justinfagnani, line 288: Don't worry about this, but I thought about it since you're subclassing and adding components... another possible UI here is an ⓧ control on the right of the tab and making the filename editable inline with the tab.

src/playground-tab-bar.ts
```diff
@@ -0,0 +1,376 @@
...
+        opacity: 70%;
+        --mdc-icon-button-size: 24px;
+        --mdc-icon-size: 16px;
+      }
+
+      .menu-button:hover {
+        opacity: 100%;
+      }
+    `,
+  ] as unknown) as CSSResult;
```
justinfagnani, line 337: argh, I'm glad we fixed this in lit-next

justinfagnani (approved):

src/playground-tab-bar.ts
aomarks: I like the dropdown because it gives us a place for more actions in the future. E.g. I was thinking of "Duplicate" and "Format".

src/playground-tab-bar.ts
aomarks: Acknowledged.

PR #96, opened 2021-01-21
Only add ?module to unpkg.com URLs

aomarks:
Snowpack errors with this unrecognized param.

Also fix some z-index issues, and show the "Result" label on previews by default.

### DIFF
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -76,6 +76,10 @@ export class PlaygroundCodeEditor extends LitElement {
         display: block;
         font-family: var(--playground-code-font-family, monospace);
         font-size: var(--playground-code-font-size, unset);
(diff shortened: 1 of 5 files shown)


src/playground-code-editor.ts
```diff
@@ -76,6 +76,10 @@ export class PlaygroundCodeEditor extends LitElement {
         display: block;
         font-family: var(--playground-code-font-family, monospace);
         font-size: var(--playground-code-font-size, unset);
+        /* CodeMirror uses z-indexes up to 6 to e.g. place scrollbars above the
+        code area. However, this can create undesirable stacking effects with
+        the rest of the page. Force a new stacking context. */
+        isolation: isolate;
```
justinfagnani, line 82: TIL!

justinfagnani (approved):

PR #99, opened 2021-01-26
Latest-wins behavior for projectSrc vs files property

aomarks:
Previously, `files` always won over `projectSrc`. Now, the most recently set wins.

This came up for the lit playground, because we need to toggle between the two ways, as the user moves between static examples and base64 URLs.

Also:

- Refactor `<playground-ide>` so that its `files` getter/setter now directly delegates to `<playground-project>` instead of keeping its own copy, to eliminate out-of-sync errors, and a buggy property-setting loop.
- Emit `compileStart` and `compileDone` events, so that the preview can start indicating that a reload is happening even before compilation completes.
- Reduce the reload-after-typing time from 500ms to 300ms to feel more responsive.

### DIFF
--- a/src/playground-ide.ts
+++ b/src/playground-ide.ts
@@ -153,9 +153,60 @@ export class PlaygroundIde extends LitElement {
 
   /**
    * A document-relative path to a project configuration file.
+   *
+   * When both `projectSrc` and `files` are set, the one set most recently wins.
+   * Slotted children win only if both `projectSrc` and `files` are undefined.
    */
-  @property({attribute: 'project-src'})
-  projectSrc?: string;
+  @property({attribute: 'project-src', hasChanged: () => false})
+  get projectSrc(): string | undefined {
+    // To minimize synchronization complexity, we delegate the `projectSrc` and
+    // `files` getters/setters directly to our <playground-project>. The only
+    // case we need to handle is properties set before upgrade or before we
+    // first render the <playground-project>.
+    //
+    // TODO(aomarks) Maybe a "delegate" decorator for this pattern?
+    const project = this._project;
+    if (project) {
+      return project.projectSrc;
+    } else {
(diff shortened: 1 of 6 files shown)


src/playground-ide.ts
```diff
@@ -153,9 +153,60 @@ export class PlaygroundIde extends LitElement {
 
   /**
    * A document-relative path to a project configuration file.
+   *
+   * When both `projectSrc` and `files` are set, the one set most recently wins.
+   * Slotted children win only if both `projectSrc` and `files` are undefined.
    */
-  @property({attribute: 'project-src'})
-  projectSrc?: string;
+  @property({attribute: 'project-src', hasChanged: () => false})
```
justinfagnani, line 160: can you add a comment on the `hasChanged: () => false` option?

src/playground-ide.ts
```diff
@@ -153,9 +153,60 @@ export class PlaygroundIde extends LitElement {
...
+      return project.projectSrc;
+    } else {
+      // We haven't rendered yet.
+      return this._projectSrcSetBeforeRender;
+    }
+  }
+
+  set projectSrc(src: string | undefined) {
+    const project = this._project;
+    if (project) {
```
justinfagnani, line 179: can `project` ever go from being defined to undefined?

justinfagnani (approved):
nice PR!

src/playground-ide.ts
aomarks: Done

```
// Note we set `hasChanged: () => false` because we don't need to trigger
// `update` when this property changes. (Why be a lit property at all?
// Because we want [1] to respond to atribute changes, and [2] to inherit
// property values set before upgrade)
```

src/playground-ide.ts
aomarks: No

PR #108, opened 2021-02-18
Make Playground importable from Skypack

aomarks:
### Skypack

Skypack is now fully working: https://glitch.com/edit/#!/teal-splashy-moose

The first problem and solution is described at https://github.com/PolymerLabs/playground-elements/issues/107. Basically the `playground-typescript-worker.js` and `playground-worker-proxy.html` files don't get served from the `mode=imports/optimized` Skypack bundle URL space, so they 404'd. Now we automatically detect when we're on Skypack, and switch to raw mode for those files.

The second problem (which also affected unpkg) was that worker scripts can only be loaded from same-origin URLs. Now we workaround by detecting different origin, manually fetching the worker script, and creating a blob URL instead.

### Unpkg

Unpkg is *partially* working now: https://glitch.com/edit/#!/scratch-iridescent-blob

There were a few problems on our side for unpkg:

1. Nullish coalescing `??` operator still isn't supported. Just emitting es2019 for now.
2. Encode the worker proxy session ID in the proxy iframe URL hash, instead of in a query param, because unpkg strips query params.
3. Removed an import of `@material/mwc-list-item`, because it was already being imported via a different URL by `@material/mwc-list`.

This lets us import`project`, `file-editor`, and `preview`. However, we still can't import `ide` or `tab-bar`. This is because it's not currently possible to import both `@material/mwc-button` and `@material/mwc-tab` at the same time from unpkg. I have a fix out for that at https://github.com/material-components/material-components-web-components/pull/2177.

### Misc

- Fix bug where `--playground-tab-bar-foreground` didn't apply to icon buttons in the tab bar.
- Add a higher resolution preview image to the README.
- Fix missing `type="module"` in the README.

Fixes https://github.com/PolymerLabs/playground-elements/issues/107


aomarks:
Found and fixed another problem. I didn't notice before that the typescript worker wasn't starting up on either skypack or unpkg because of the same-origin policy. Now we detect the different origin, `fetch` the script manually, and create a `Blob` url instead.

src/playground-file-system-controls.ts
```diff
@@ -22,11 +22,15 @@ import {
 } from 'lit-element';
 import {nothing} from 'lit-html';
 
-import '@material/mwc-menu/mwc-menu-surface.js';
+// Note despite usual best practices, we should _not_ import
+// @material/mwc-list-item directly, because @material/mwc-list already imports
+// it, and this causes a duplicate registration error on unpkg.com because of
+// redirects.
```
justinfagnani, line 28: even if everything resolve to the same version? I would have thought with ?module that the final URL is written into the import

src/playground-project.ts
```diff
@@ -50,10 +54,38 @@ const generateUniqueSessionId = (): string => {
...
+    const text = await resp.text();
+    const blob = new Blob([text], {type: 'application/javascript'});
+    const blobUrl = URL.createObjectURL(blob);
+    deferredWorkerBlobUrl.resolve(blobUrl);
+    setTimeout(() => {
+      // The worker script is 2.7MB. Only cache it for a short while in case a
+      // bunch of playgrounds are starting up at the same time.
+      deferredWorkerBlobUrl = undefined;
+      URL.revokeObjectURL(blobUrl);
+    }, 1000);
```
justinfagnani, line 85: Do you need to do this? afaik, `fetch()` by default uses the HTTP cache along with normal handling of Cache-Control, 304s, etags and such. https://developer.mozilla.org/en-US/docs/Web/API/Request/cache

If you do need this, add a comment on why the HTTP cache isn't sufficient, and I'd test on slow network and CPU and maybe bump to 5-10s if necessary.

src/shared/util.ts
```diff
@@ -16,3 +16,17 @@ export const endWithSlash = (s: string) => (s.endsWith('/') ? s : s + '/');
 
 export const getRandomString = () =>
   crypto.getRandomValues(new Uint32Array(1))[0].toString(32);
+
+/**
+ * If the given URL object is a Skypack URL, perform an in-place update that
+ * switches from optimized mode to raw mode.
```
justinfagnani, line 22: is there a Skypack bug you can reference?

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "composite": true,
     "incremental": true,
-    "target": "esnext",
+    "target": "es2019",
```
justinfagnani, line 5: ES2020 to pick up optional chaining and nullish coalescing? That's what we're going to ship Lit 2 with

justinfagnani (approved):

src/playground-project.ts
aomarks: Now revoking immediately.

src/shared/util.ts
aomarks: Added link to our bug.

PR #111, opened 2021-02-23
Publish a Playground stylesheet for SSR

aomarks:
**Note**: I recommend looking at each commit separately. The first is the pristine CodeMirror styles.

We need to be able to import our customized CodeMirror styles into a different scope when doing SSR. This PR adds `playground-styles.css` and `playground-styles.js` files for doing that.

Also:

- Replace stylesheet generation with a manually transformed version of the CodeMirror styles. Easier to understand and maintain.
- Add some custom properties
- Change some defaults (slightly less dense, some more sane inheritance)


scripts/theme-generator.js
```diff
@@ -6,17 +6,13 @@
...
-//   1) src/_codemirror/codemirror-styles.ts : A transformed version of the default
-//      CodeMirror stylesheet, which replaces fixed default colors with custom
-//      properties.
-//
-//   2) themes/*.css : A CSS file for each of the standard CodeMirror themes, which
+//   1) themes/*.css : A CSS file for each of the standard CodeMirror themes, which
 //      expresses that theme in terms of CSS custom properties.
 //
-//   3) src/themes/*.css.ts : A JavaScript module for each file in #2 that exports
+//   2) src/themes/*.css.ts : A JavaScript module for each file in #2 that exports
```
sorvell, line 12: in #1?

sorvell (approved):

scripts/theme-generator.js
aomarks: #2 is the TS module, #1 is the css file

PR #113, opened 2021-03-03
Fix fold/hide infinite loop bug. Add content/contentType properties. Shorter license headers.

aomarks:
- Fix infinite loop bug that happened when switching between two files that both had fold/hide regions.
 
- Add content and contentType optional properties to JSON manifest. This can be used to more efficiently serve a playground project, with one request instead of N.

- Use short SPDX style headers.


kevinpschaaf (approved):
LGTM, I'll take your word on breaking out of the codemirror stack! :P

PR #118, opened 2021-03-24
Add extends field to JSON configs

aomarks:
You can now set the `extends` property of a JSON config to another config file URL, and they'll be fetched and merged recursively. Useful for when we have a base config for a site that we want to inherit across a bunch of playgrounds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Set the `project-src` attribute or `projectSrc` property to a JSON file with for
 | `files.contentType` | Optional MIME type of the file. If omitted, type is taken from either the `fetch` response `Content-Type` header, or inferred from the filename extension when `content` is set. |
 | `files.label`       | Optional label for display in `playground-tab-bar`. If omitted, the filename is displayed.                                                                                       |
 | `files.hidden`      | If `true`, the file won't be visible in `playground-tab-bar`.                                                                                                                    |
-| `extends`           | Optional URL to another JSON config file to extend from. Configurations are deeply merged.                                                                                       |
+| `extends`           | Optional URL to another JSON config file to extend from. Configs are deeply merged. URLs are interpreted relative to the URL of each extendee config.                            |
 
 ```html
 <playground-ide project-src="/path/to/my/project.json"> </playground-ide>
--- a/README.md
+++ b/README.md
@@ -163,22 +163,16 @@ There are 3 ways to specify the files of a playground project:
 
 ### Option 1: Inline scripts
 
-Add one or more `<script type="sample/..." filename="...">` tags as children of
-your `<playground-ide>` or `<playground-project>`. The following attributes are
-available:
+Add one or more `<script>` tags as children of
+your `<playground-ide>` or `<playground-project>`, using the following attributes:
 
-- `type`: Required filetype. Valid options: `sample/html`, `sample/css`,
-  `sample/js`, `sample/ts`, `sample/json`, `sample/importmap`.
-
-- `filename`: Required filename.
-
-- `label`: Optional label for display in `playground-tab-bar`. If omitted, the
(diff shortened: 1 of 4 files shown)


src/playground-project.ts
```diff
@@ -519,39 +520,76 @@ export class PlaygroundProject extends LitElement {
...
-          content,
-          contentType,
-        };
+const fetchProjectConfig = async (
+  url: string
+): Promise<{files: Array<SampleFile>; importMap: ModuleImportMap}> => {
+  const mergedFilePromises = new Map<string, Promise<FileOptions>>();
+  const mergedImportMap = {imports: {}};
+
+  const fetchAndMerge = async (configUrl: string) => {
```
e111077, line 529: Can this closure be pulled out of the class or file and turned into a normal function? It doesn't seem to have element state side effects and passing in mergedFilePromises or mergedImportMap seems to be an okay solution.

My reasoning is that I think it's a bit less readable with it inlined like this. And to make it a bit more readable despite it mutating some of the arguments, you can make it return the same instance of `mergedFilePromises` and `mergedImportMap` in something like this

`{mergedFilePromises, mergedImportMap} = await fetchAndMerge(url, mergedFilePromises, mergedImportMap);`

Alternatively you can make it immutable but that's probably too much churn for "correctness"

README.md
```diff
@@ -210,23 +204,18 @@ Be sure to escape closing `</script>` tags within your source as `<&lt;script>`.
...
+| `files.content`     | Optional text content of the file. If omitted, a `fetch` is made to retrieve the file by filename, relative to the manifest URL.                                                 |
+| `files.contentType` | Optional MIME type of the file. If omitted, type is taken from either the `fetch` response `Content-Type` header, or inferred from the filename extension when `content` is set. |
+| `files.label`       | Optional label for display in `playground-tab-bar`. If omitted, the filename is displayed.                                                                                       |
+| `files.hidden`      | If `true`, the file won't be visible in `playground-tab-bar`.                                                                                                                    |
+| `extends`           | Optional URL to another JSON config file to extend from. Configurations are deeply merged.                                                                                       |
```
e111077, line 218: Is this relative or absolute and relative to what? (is this even an applicable question I'm asking?)

e111077 (changes requested):

src/playground-project.ts
```diff
@@ -519,39 +520,76 @@ export class PlaygroundProject extends LitElement {
...
+          filename,
+          (async () => {
+            const resp = await fetch(new URL(filename, configUrl).href);
+            info.contentType = resp.headers.get('Content-Type') ?? undefined;
+            info.content = await resp.text();
+            return info;
+          })()
+        );
+      } else {
+        info.contentType ??= typeFromFilename(filename) ?? 'text/plain';
```
e111077, line 563: ?? ??? ?? ? ???? ?? ????

README.md
aomarks: Updated the comment, is it clear? Was surprisingly hard to write a sentence that specifies it. I think it's the intuitive behavior (the URL is relative to the place where you wrote the URL).

src/playground-project.ts
aomarks: Good call, I refactored it into a single recursive function, hopefully better?

src/playground-project.ts
aomarks: ¿߄ᕍᣢᣑᕉ?

e111077 (approved):
LGTM

e111077 (commented):
wait one sec gotta double check something!

e111077 (changes requested):
block merge

e111077 (approved):
yeah, I like this implementation much better

PR #120, opened 2021-03-28
Show basic TypeScript errors

aomarks:
- Syntactic and other *basic* TypeScript errors are now shown with red underlines and a tooltip on hover.

- We currently can't show any errors that depend on typings, because we don't fetch them (even the standard libs)! This will require some more work -- tracking separately at https://github.com/PolymerLabs/playground-elements/issues/119 and will send as followup.

- It's using the Language Server Protocol `Diagnostic` type as the interface. Should be good for other diagnostic sources we might add in the future.

- Sadly, we're currently displaying a `border-bottom: 2px dashed red`, instead of `text-decoration: red wavy underline`, because `wavy` doesn't render *anything* for single characters (!): https://bugs.chromium.org/p/chromium/issues/detail?id=668042). Could add a custom implementation of wavy underlines as a followup.

- Try it out at https://polymerlabs.github.io/playground-elements/

Part of https://github.com/PolymerLabs/playground-elements/issues/67

### DIFF
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -155,6 +155,7 @@ export class PlaygroundCodeEditor extends LitElement {
   };
 
   private _resizeObserver?: ResizeObserver;
+  private _resizing = false;
   private _valueChangingFromOutside = false;
   private _ignoreValueChange = false;
   private _hideOrFoldRegionsActive = false;
@@ -222,13 +223,13 @@ export class PlaygroundCodeEditor extends LitElement {
     // sometimes be missing, but typing in the editor will fix it.
     if (typeof ResizeObserver === 'function') {
       this._resizeObserver = new ResizeObserver(() => {
-        // Temporarily disconnect to avoid loops, because CodeMirror might
-        // resize itself on refresh.
-        this._resizeObserver?.disconnect();
+        if (this._resizing) {
+          // Don't get in a resize loop.
+          return;
+        }
+        this._resizing = true;
         this._codemirror?.refresh();
-        requestAnimationFrame(() => {
-          this._resizeObserver?.observe(this);
-        });
+        this._resizing = false;
       });
       this._resizeObserver.observe(this);
     }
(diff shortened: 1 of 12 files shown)


src/playground-code-editor.ts
```diff
@@ -114,6 +63,29 @@ export class PlaygroundCodeEditor extends LitElement {
         /* Pretty much any color from the theme is good enough. */
         color: var(--playground-code-keyword-color, #770088);
       }
+
+      .diagnostic {
+        /* It would be nice to use "text-decoration: red wavy underline" here,
+           but unfortunately it renders nothing at all for single characters.
+           See https://bugs.chromium.org/p/chromium/issues/detail?id=668042. */
+        border-bottom: var(--playground-error-border, 2px red dashed);
```
justinfagnani, line 71: Q: this seems to layout fine, but I wondered if `outline` was the correct property to use.?

src/playground-code-editor.ts
```diff
@@ -203,6 +189,9 @@ export class PlaygroundCodeEditor extends LitElement {
           case 'pragmas':
             this._applyHideAndFoldRegions();
             break;
+          case 'diagnostics':
+            this._showDiagnostics();
+            break;
```
justinfagnani, line 194: I'd love to have an `@observe` decorator or similar for this pattern...

src/typescript-worker/playground-typescript-worker.ts
```diff
@@ -425,3 +443,62 @@ class WorkerLanguageServiceHost implements ts.LanguageServiceHost {
     return '__lib.d.ts';
   }
 }
+
+/**
+ * Convert a diagnostic from TypeScript format to Language Server Protocol
+ * format.
+ */
+function makeLspDiagnostic(tsDiagnostic: ts.Diagnostic): lsp.Diagnostic {
```
justinfagnani, line 451: Interesting. You didn't see a function like this in typescript itself?

justinfagnani (approved):
very nice!

PR #124, opened 2021-04-02
Fetch types and show semantic errors

aomarks:
- Squiggles are now shown for type errors (previously just syntax errors).
- Walks the module graph and fetches `.d.ts` files for each import.
- Places them into the TypeScript in-memory file system with the `node_modules/` paths that we anticipate TypeScript will search for.
- Follows `package.json` `main` field where necessary.
- Uses the existing module resolver, including import maps, to decide where to fetch from.
- Loads standard libs too.
- Since fetching and type checking is slow, it's done asynchronously. The JS emit and syntax errors come just as fast as before, while the new semantic errors show a bit later.
- Squiggles are now shown as red wavy lines instead of dashed straight line.

Live at https://polymerlabs.github.io/playground-elements/

Fixes https://github.com/PolymerLabs/playground-elements/issues/119

image


src/typescript-worker/module-resolver.ts
```diff
@@ -0,0 +1,105 @@
+/**
```
aomarks, line 1: Note this file is not new, just factored out of the worker.

justinfagnani:
A small nit on styling - but rendering multiple consecutive spans with a dashed underlying is not rendered well by Chrome:

Screen Shot 2021-04-02 at 9 20 50 AM

Dotted looks _slightly_ better IMO:

Screen Shot 2021-04-02 at 9 21 04 AM

The CodeMirror lint demo has a more traditional squiggly line - though it's a bit pixelated: https://codemirror.net/demo/lint.html

The use a background image to draw the underline, and it seems to line up well, at least for the particular font in the demo
https://github.com/codemirror/CodeMirror/blob/master/addon/lint/lint.css#L38

configurator/project/project.json
```diff
@@ -2,5 +2,18 @@
...
+    "imports": {
+      "lit": "https://cdn.skypack.dev/lit@2.0.0-pre.2/",
+      "lit/": "https://cdn.skypack.dev/lit@2.0.0-pre.2/",
+      "@lit/reactive-element": "https://cdn.skypack.dev/@lit/reactive-element",
+      "@lit/reactive-element/": "https://cdn.skypack.dev/@lit/reactive-element/",
+      "lit-element": "https://cdn.skypack.dev/lit-element@3.0.0-pre.4",
+      "lit-element/": "https://cdn.skypack.dev/lit-element@3.0.0-pre.4/",
+      "lit-html": "https://cdn.skypack.dev/lit-html@2.0.0-pre.7",
+      "lit-html/": "https://cdn.skypack.dev/lit-html@2.0.0-pre.7/",
+      "typescript/": "https://unpkg.com/typescript@4.2.3/"
```
justinfagnani, line 16: do all projects need this now?

src/typescript-worker/playground-typescript-worker.ts
```diff
@@ -65,10 +68,35 @@ const workerAPI: TypeScriptWorkerAPI = {
...
+    // don't wait for them).
     const moduleResolver = new ModuleResolver(importMap);
-    const loadedFiles = await loadFiles(files, moduleResolver);
+    const loadedFiles = new Map<string, FileRecord>();
+    const typesFetcher = new TypesFetcher(moduleResolver);
+    const inputFiles = files
+      .filter((file) => isTsOrJsFile(file.name))
+      .map((file) => ({
+        file,
+        // TODO: include session/scope id?
```
justinfagnani, line 85: can you expand? Why might you want to do that?

src/typescript-worker/playground-typescript-worker.ts
```diff
@@ -90,34 +120,50 @@ const workerAPI: TypeScriptWorkerAPI = {
...
+        (fileName: string, data: string) => {
+          emittedFiles.set(fileName, data);
+        },
+        undefined,
+        undefined,
+        transformers
+      );
+    }
+
+    (async () => {
```
justinfagnani, line 143: add comment?

src/typescript-worker/playground-typescript-worker.ts
```diff
@@ -90,34 +120,50 @@ const workerAPI: TypeScriptWorkerAPI = {
...
+        // TypeScript is going to look for these files as paths relative to our
+        // source files, so we need to add them to our filesystem with those
+        // URLs.
+        const url = new URL(`node_modules/${specifier}`, self.origin).href;
+        loadedFiles.set(url, {
+          status: 'resolved',
+          content,
+        });
+      }
+      for (const {file, url} of inputFiles) {
```
justinfagnani, line 155: Do you need to bail based on the compileID here too?

src/typescript-worker/types-fetcher.ts
```diff
@@ -0,0 +1,291 @@
...
+    Deferred<Result<string, number>>
+  >();
+
+  constructor(moduleResolver: ModuleResolver) {
+    this._moduleResolver = moduleResolver;
+  }
+
+  /**
+   * Start fetching type definitions for all bare module specifiers in the given
+   * TypeScript source text. Relative module specifiers are ignored.
```
justinfagnani, line 31: >Relative module specifiers are ignored

relative specifiers are followed for transitive imports, yes? A dependency imported via a bare specifier might relatively import a same-package module that has typings.

src/typescript-worker/types-fetcher.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  /**
+   * Start fetching type definitions for a built-in TypeScript lib, like "dom"
+   * or "esnext".
+   *
+   * This function returns immediately, but begins an asynchronous walk of the
+   * <reference> graph. To access the results, await {@link getTypings}.
+   */
+  addLibTypings(lib: string): void {
```
justinfagnani, line 57: Are the URL for these based on the `"typescript"` entry in the import map?

src/typescript-worker/types-fetcher.ts
```diff
@@ -0,0 +1,291 @@
...
+    if (resp.status === 200) {
+      result = {result: await resp.text()};
+    } else {
+      result = {error: resp.status};
+    }
+    deferred.resolve(result);
+    return result;
+  }
+
+  private async _fetchPackageJson(
```
justinfagnani, line 231: do you want to cache these? Are you kick off multiple fetches for package.json on `lit`, `lit/decorators.js`, `lit/directives/repeat.js`, etc?

aomarks: Yeah, it is actually cached via `_fetchAsset`

src/typescript-worker/types-fetcher.ts
aomarks: Yep, relative specifiers are supported, just not from the top level. We don't care about them in the top level, because any relative path there would be a project file, which we are already handling. But if a project file imports a bare module, which imports a relative path, it will be handled correctly.

src/typescript-worker/playground-typescript-worker.ts
aomarks: We could support cancellation here as an optimization, but it's not a correctness issue. The callback on the project side will reject a stale result.

I think it's probably not a big deal for now. Most of the time is spent fetching files, which will probably all be needed for a subsequent compile anyway, so it mostly won't be that big a waste of time.

aomarks (commented):
> The use a background image to draw the underline, and it seems to line up well, at least for the particular font in the demo

Good call, done. I wanted to use `red wavy underline`, but unfortunately it displays nothing when there is only one character (https://bugs.chromium.org/p/chromium/issues/detail?id=668042). I cribbed this `png` for now, maybe an improvement would be to use SVG and apply a custom property color to it.

src/typescript-worker/playground-typescript-worker.ts
aomarks: Done.

justinfagnani (approved):

PR #129, opened 2021-04-19
Tab inserts 2 spaces instead of tab character

aomarks:


e111077 (approved):

rwalle:
I noticed that indenting is not working properly -- tabs always insert 1 space instead of 2 -- and tracked it down to this change. The implementation is not correct -- `Array(cm.getOption('indentUnit') ?? 2).join(' ')` gives you `' '` (one space) instead of `'  '` (two spaces). You would need to use `Array(n+1)` or do `Array(2).fill(' ').join('')`

PR #136, opened 2021-04-23
Document Playground sandbox security

aomarks:


aomarks:
Actually going to think about this some more -- I think it might make sense to set the sandbox URL to unpkg.com by default, because that should work and will be secure out of the box by default.

PR #137, opened 2021-04-25
Sandbox security

aomarks:
Untrusted code now executes under the `unpkg.com` origin by default, instead of under the component-host origin. This is a more secure default, because `unpkg.com` doesn't have access to cookies, privileged APIs, etc., and can't access `window.parent`.

It also simplifies bundling, because the dynamically-loaded `playground-service-worker.js` and `playground-service-worker-proxy.html` files no longer need to be copied into the bundled output directory by default.

We enforce that version numbers are synchronized between the local components and the `unpkg.com` URL with a `prepack` (and hence pre-publish) script.

It might seem a little unusual to have a core dependency on `unpkg.com` baked-in like this, but note that since we already depend on `unpkg.com` for bare module resolution, it doesn't really constitute a new potential failure point for most users. For users who would like to host their own secure playground origin (as we do on lit.dev), there is now a new "sandbox security" section in the README.


src/playground-ide.ts
```diff
@@ -21,7 +21,7 @@ import './playground-file-editor.js';
 import './playground-preview.js';
 import {PlaygroundProject} from './playground-project.js';
 import {ProjectManifest} from './shared/worker-api.js';
-import {forceSkypackRawMode} from './shared/util.js';
```
e111077, line 24: For you use the fn anywhere else? If not don't forget to delete it

e111077 (changes requested):

src/playground-ide.ts
aomarks: Yep, it's also used in `playground-project.ts` to resolve the typescript web worker when imported from skypack.

e111077 (approved):

PR #142, opened 2021-05-02
Be more responsive when compiles are fast

rictic:
Currently, due to the use of a trailing edge debouncer, it takes a
minimum of 300ms between the user making an edit and the result
displaying on screen. However, in many cases we can complete an
edit -> compile -> display pipeline in just a few dozen ms.

This – combined with displaying an attention-getting animated progress
bar for a full half second – makes the playground feel less
responsive than it could be.

This change makes some changes to optimize the UX for fast compiles.

- Change from a trailing edge debouncer to leading edge, so the
  system responds as fast as possible to the first change made.
- If there is a pending compile, start it the moment the previous
  compile finishes. This is somewhat more wasteful, particularly
  for slow compiles, but displays more intermediate states, reducing
  the delta between the displayed code and the displayed preview.
- Removes the minimum display time for the progress bar. Currently
  the progress bar is often the most prominent part of the preview,
  since it's animated, and is displayed for a minimum of half a second.
  I think it's better for the focus to be on the preview contents,
  with the progress bar only displaying when necessary to
  indicate that a preview is loading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Error tooltips now follow `--playground-code-font-family`.
+- Optimized the UX for fast compile and display by switching to a leading edge
+  debouncer and eliminating the minimum display time for the loading bar.
 
 ## [0.9.2] - 2021-04-26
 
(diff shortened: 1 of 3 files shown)


aomarks (approved):
Tried it out, feels great!

PR #158, opened 2021-05-18
Pin comlink dependency

aomarks:
By default, for security reasons, the playground service worker is served from unpkg.com even when components are installed locally through NPM.

The service worker includes a bundled copy of comlink, since Firefox hasn't yet shipped module imports from workers.

So, a breakage occured because of a patch release to comlink, which meant the locally installed components had a different version of comlink than the service worker side.

(I think GoogleChromeLabs/comlink#475 was the breaking change, since it changed the values of some enums).

This change pins the comlink version to prevent this from happening again.

Also bumps the other dependencies.

This *should* fix https://github.com/PolymerLabs/playground-elements/issues/156 and https://github.com/PolymerLabs/playground-elements/issues/157


src/playground-code-editor.ts
```diff
@@ -518,7 +518,7 @@ export class PlaygroundCodeEditor extends LitElement {
       case 'css':
         return 'css';
     }
-    return null;
+    return undefined;
```
rictic, line 521: Needed for updated typings?

rictic (approved):

src/playground-code-editor.ts
aomarks: Yeah, updated codemirror typings.

PR #159, opened 2021-05-25
Replace mwc-tab-bar with custom tab bar

aomarks:
**Note to reviewer**: I highly recommend looking at each commit separately.

Reasons:

- We currently have some very hacky code, including a subclass of mwc-tab, to support context menus. Our own implementation can do this much more straightforwardly: https://github.com/PolymerLabs/playground-elements/blob/271423b61591c1beceb2da44fc74d3c176eb7fdd/src/playground-tab-bar.ts#L290

- We want to support re-ordering, and that would be difficult with mwc-tab-bar (https://github.com/PolymerLabs/playground-elements/issues/146).

- mwc-tab-bar is broken on skypack, so playground is broken too (https://github.com/skypackjs/skypack-cdn/issues/169).

- There's a timing bug in mwc-tab-bar which we have a weird workaround for: https://github.com/PolymerLabs/playground-elements/blob/271423b61591c1beceb2da44fc74d3c176eb7fdd/src/playground-tab-bar.ts#L195

- It's smaller (119 KiB raw / 7 KiB min+brot vs 317 / 18)

Fixes https://github.com/PolymerLabs/playground-elements/issues/147


src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+          @tabchange=${this._activateTab}
+        ></slot>
+      </div>
+    `;
+  }
+
+  private _onSlotchange(event: Event) {
+    this._tabs = (
+      event.target as HTMLSlotElement
+    ).assignedElements() as PlaygroundInternalTab[];
```
e111077, line 99: not having to support IE is amazing! assignedElements is not in the polyfill only assignedNodes

e111077, line 99: How "custom" do you want this to be? like only for your use case? might be a problem if someone tries to use it and places another element as a child? Might be good to filter? In MWC list we assume the interface of any component with the attribute `[mwc-list-item]` is `ListItem` and filter out other things like list separators.

Close comment if you don't expect to use this outside of the bounds of PgE's.

If you do decide to go with this, make sure you change references to `localName` below

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+    let newActive;
+    // Manage the idx and active properties on all tabs. The first tab that
+    // asserts it is active wins.
+    for (let i = 0; i < this._tabs.length; i++) {
+      const tab = this._tabs[i];
+      tab.index = i;
```
e111077, line 105: What happens if the tab is not upgraded by the time slotchange runs? I'd recommend `setAttribute('index', i)`

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+      if (newActive !== undefined) {
+        tab.active = false;
```
e111077, line 107: same as above. Might also want to `removeAttribute`

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  private _tabs: PlaygroundInternalTab[] = [];
+  private _active: PlaygroundInternalTab | undefined = undefined;
+
+  render() {
+    return html`
+      <div role="tablist">
+        <slot
+          @slotchange=${this._onSlotchange}
```
e111077, line 87: We are assuming shadow DOM on, right?

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+        // not have reflected to the other yet.
+        newActive = tab;
+      }
+    }
+    this.active = newActive;
+  }
+
+  private _activateTab(event: Event) {
+    const tab = this._findEventTab(event);
+    if (tab === undefined || tab === this.active) {
```
e111077, line 120: I tend to find it's a bit more robust if the tab "requests" active rather than set it on itself (outside of initialization) and the bar acts as a controller between it and its siblings and is the one that actually sets active on the child. But I think this component is simple enough that it's probs fine to not do this. Resolve if you don't feel like doing it.

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+      return;
+    }
+    this.active = tab;
+    tab?.scrollIntoView({behavior: 'smooth'});
```
e111077, line 124: Might this cause issues on page load? Like scroll to your tabs no matter where you are on the page? How does this affect a docs page with multiple playground IDEs on the page?

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  private async _onKeydown(event: KeyboardEvent) {
```
e111077, line 127: Missing Home and End

https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html#kbd_label

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+    const oldIdx = this.active?.index ?? 0;
+    let newIdx = oldIdx;
+    if (event.key === 'ArrowLeft') {
+      if (oldIdx > 0) {
```
e111077, line 131: should consider wrapping when `oldIdx === 0` and at other end

https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html#kbd_label

src/internal/tab.ts
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * The 0-indexed position of this tab within its <playground-internal-tab-bar>.
+   *
+   * Note this property is managed by the containing <playground-internal-tab-bar> and
+   * should not be set directly.
+   */
+  index = 0;
+
+  update(changes: PropertyValues) {
+    if (changes.has('active')) {
```
e111077, line 85: What's the reasoning for doing this in update and not updated? is tabindex or aria-selected important to tab-bar at this point?

(Not a leading question, I genuinely don't know)

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+      );
+    }
+  }
+
+  private _tabs: PlaygroundInternalTab[] = [];
+  private _active: PlaygroundInternalTab | undefined = undefined;
+
+  render() {
+    return html`
+      <div role="tablist">
```
e111077, line 85: w3c says you want aria-label here.

Since you know the exact purpose for this tab bar you likely can hard code this.

src/internal/tab-bar.ts
```diff
@@ -0,0 +1,168 @@
...
+          @slotchange=${this._onSlotchange}
+          @click=${this._activateTab}
+          @keydown=${this._onKeydown}
+          @tabchange=${this._activateTab}
+        ></slot>
+      </div>
+    `;
+  }
+
+  private _onSlotchange(event: Event) {
```
e111077, line 96: does this fire on disconnect. e.g. <tab-bar><lots of tabs></tab-bar>

`tabBar.parentElement.removeChild(tabBar)`

If so, you may want to prevent this from running then because then switching pages in a SPA might become costly because of this.

e111077 (changes requested):

e111077:
RIP Shadow DOM and aria-controls

src/internal/tab-bar.ts
aomarks: Oh yeah, good point!

Replaced with a  custom implementation. I think this new approach is also better because it also tries to make sure adjacent tabs are at least 48px visible, otherwise they can disappear so that it's hard to navigate to the next/prev one.

src/internal/tab-bar.ts
e111077: well we don't have a use for it now, so I think it's fine

src/internal/tab-bar.ts
e111077: IIRC shady you cannot add an event to the slot itself. it bubbles so it's typically placed on parent

src/internal/tab-bar.ts
e111077: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html#rps_label

e111077 (approved):

PR #162, opened 2021-06-04
Fix invalid default padding value

aomarks:
Small syntax typo from https://github.com/PolymerLabs/playground-elements/pull/161


AndrewJakubowicz (approved):
Looks good to me!

PR #163, opened 2021-06-04
Refactor playground worker API to be streaming

aomarks:
***Reviewers should definitely look at each commit separately!***

- Defines a streaming API for composing *builders* into a pipeline within the web worker. Each builder receives and returns an async iterator of results. Results are files or diagnostics.

- For now, we just have the same TypeScript builder, slightly refactored to use the new API. In subsequent PRs coming soon I'm planning to add:
  - A builder that extracts inline JS from HTML so that we can process that JS too.
  - A dedicated bare module transform builder that uses https://github.com/guybedford/es-module-lexer, respects `package.json` dependencies, fetches external dependencies and transforms those too, so that we can correctly canonicalize all import specifiers.

- Replaces the crude "slow" vs "fast" diagnostics distinction we had before. Now diagnostics just stream whenever they are found.

- We now serve each asset as soon as it is ready, instead of waiting for the entire build to finish before loading. Most notably this means the `index.html` can start rendering concurrently with the TypeScript compile. The GET request for each asset blocks until it emerges from the stream (and 404s if the stream ends without that asset being found).

- Adds tests that check the build output directly (previously we only had full e2e tests that didn't couldn't easily look at build outputs in detail).

Related to https://github.com/PolymerLabs/playground-elements/issues/66


src/playground-project.ts
```diff
@@ -464,7 +464,7 @@ export class PlaygroundProject extends LitElement {
     port.start();
   }
 
-  private async _getFile(name: string): Promise<SampleFile | undefined> {
+  private async _getFile(name: string): Promise<SampleFile | HttpError> {
```
justinfagnani, line 459: Seems weird to have an error in the completion type. Is it here because the rejection type isn't generic?

src/playground-project.ts
```diff
@@ -465,74 +457,39 @@ export class PlaygroundProject extends LitElement {
...
-    this.dispatchEvent(new CustomEvent('diagnosticsChanged'));
-  }
-
-  private _saveTimeoutId?: ReturnType<typeof setTimeout> = undefined;
-
-  private _clearSaveTimeout() {
-    if (this._saveTimeoutId !== undefined) {
-      clearTimeout(this._saveTimeoutId);
-      this._saveTimeoutId = undefined;
+    workerApi.compileProject(
```
justinfagnani, line 483: Eventually it'd be nice to enable incremental builds for build-as-you-type and generally taking less time> Especially as deps won't change.

justinfagnani (approved):

src/playground-project.ts
aomarks: Yeah that's the main reason.

It also feels conceptually like HTTP errors here are in the normal flow. We're essentially implementing an HTTP server, and getting e.g. a 404 because of an import of a missing project file is something that will fairly often happen due to the user's input. So, getting an exception here would be reserved for some unanticipated internal error.

src/playground-project.ts
aomarks: Yep would like to tackle that sometime, after all this refactoring is done.

PR #165, opened 2021-06-04
Add dedicated bare module transformer and support dynamic import

aomarks:
- Uses [es-module-lexer](https://github.com/guybedford/es-module-lexer) to add a dedicated transform just for bare modules, and removes the bare module transform from the TypeScript builder.

- I had to make a small patch to `es-module-lexer` to make it compatible with workers (https://github.com/guybedford/es-module-lexer/pull/70). That's not released yet, so for now I am just patching the source in Rollup.

- While this is probably a bit slower to transform `.ts` files (because there's an extra parse [or lex I guess]), in a follow up PR I'm planning to add support for transforming all transitive *dependencies* with this builder too. I think it will be common to have far more *dependency* modules than project modules (e.g. if you import lit), and this lexer should be much faster than TypeScript (it uses Wasm and is very minimal since it only cares about imports/exports), so I think average build time will improve.

- I'd like to make TypeScript optional at some point, since it is the largest chunk of code by far, so it will be nice to have a small dedicated builder just for doing bare module transforms. `es-module-lexer` by itself is 4.3 KiB min/brot.

- Adds support for bare module specifiers in dynamic imports (fixes https://github.com/PolymerLabs/playground-elements/issues/27).

- This means `.js` files are no longer compiled by TypeScript, so they can't contain decorators or any TypeScript-specific syntax. (fixes https://github.com/PolymerLabs/playground-elements/issues/160).
  - I think this is an advantage, because while TypeScript *can* compile `.js`, it's relatively uncommon. I think it's more intuitive to treat `.js` files as raw JS. Currently a user might paste a snippet of TypeScript into a `.js` file (e.g. from the lit.dev docs), find that it runs just fine in the Playground, but then encounter an error when they try and reproduce the same code locally (e.g. using the download tarball button).


justinfagnani:
Looks like a new version of `es-module-lexer` is out

package.json
```diff
@@ -62,6 +62,7 @@
     "clean-css": "^5.1.2",
     "codemirror": "git+https://github.com/codemirror/CodeMirror.git#2997167571787db3da8a15e5ac65c8ddaf316f0a",
     "codemirror-grammar-mode": "^0.1.10",
+    "es-module-lexer": "=0.4.1",
```
justinfagnani, line 65: try 0.6.0

src/typescript-worker/bare-module-transformer.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  async *process(
+    results: AsyncIterable<BuildOutput> | Iterable<BuildOutput>
+  ): AsyncIterable<BuildOutput> {
+    for await (const result of results) {
+      if (
+        result.kind === 'file' &&
+        result.file.name.endsWith('.js') &&
+        result.file.content.includes('import')
```
justinfagnani, line 30: Isn't `es-module-lexer` optimized for this case? Does an extra string scan here speed anything up?

justinfagnani (approved):

package.json

src/typescript-worker/bare-module-transformer.ts
aomarks: Good point, probably just makes it worse! It was also wrong because of export from.

PR #169, opened 2021-06-10
Add a WTR plugin for simulating CDNs like unpkg.com

aomarks:
This will be useful for testing an upcoming PR that proxies unpkg.com so that we can control the module specifiers of dependencies (https://github.com/PolymerLabs/playground-elements/issues/104).

It's a plugin for @web/test-runner that simulates an NPM CDN like unpkg.com. It extends the built-in @web/test-runner server, and can be re-configured from the test suite itself between each test using a server command (https://modern-web.dev/docs/test-runner/commands/).

Usage:

```ts
import {executeServerCommand} from '@web/test-runner-commands';

const cdnData = {
  "foo": {
    "1.2.3": {
      "main": "lib/index.js",
      "files": {
        "lib/index.js": {
          "content": "console.log('hello');"
        }
      }
    }
  }
};
const cdnBaseUrl = await executeServerCommand('set-fake-cdn-data', cdnData);
// Redirects to <cdnBaseUrl>/foo@1.2.3/lib/index.js and serves its content.
const result = await fetch(new URL("foo@^1.0.0", cdnBaseUrl).href);
 ```

### DIFF
--- a/src/test/fake-cdn-plugin.ts
+++ b/src/test/fake-cdn-plugin.ts
@@ -29,14 +29,16 @@ import semver from 'semver';
  *        }
  *      }
  *    };
- *    const cdnBaseUrl = await executeServerCommand('set-fake-cdn-data', cdnData);
+ *    const {cdnBaseUrl, id} = await executeServerCommand('set-fake-cdn-data', cdnData);
  *    // Redirects to <cdnBaseUrl>/foo@1.2.3/lib/index.js and serves its content.
  *    const result = await fetch(new URL("foo@^1.0.0", cdnBaseUrl).href);
+ *    await executeServerCommand('delete-fake-cdn-data', id);
  */
 export function fakeCdnPlugin(): TestRunnerPlugin {
   const pathPrefix = '/fake-cdn/';
   let baseUrl: string | undefined;
-  let data: CdnData = {};
+  const dataMap = new Map<string, CdnData>();
+  let _nextId = 0;
 
   return {
     name: 'fake-cdn',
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/test/fake-cdn-plugin.ts [hunk 1]
```diff
@@ -0,0 +1,144 @@
...
+        semverRangeOrTag === '' || semverRangeOrTag === 'latest'
+          ? '*'
+          : semverRangeOrTag;
+      const versions = Object.keys(packageData.versions);
+      const version = semver.maxSatisfying(versions, semverRange);
+      if (!version) {
+        ctx.response.status = 404;
+        return undefined;
+      }
+      if (version !== semverRange) {
```
AndrewJakubowicz, line 112: Sanity checking my understanding of the logic here. Does this mean that a concrete version has been found in `packageData.versions` that isn't for example a semverRange?

Then this is a redirect from a range to a concrete version? I.e. a range becomes resolved into something concrete such as 1.2.3. Then we redirect calling this endpoint again with the new version. Now the version 1.2.3 would match the semverRange 1.2.3 and we'd reach line 89 where we fetch the content?

AndrewJakubowicz (approved):
The only change required is a small documentation update.

LGTM! I didn't review `parseNpmModuleSpecifier` as you've updated/tested in follow up PRs.

src/test/fake-cdn-plugin.ts
```diff
@@ -0,0 +1,144 @@
...
+ * the test suite itself between each test using a server command
+ * (https://modern-web.dev/docs/test-runner/commands/).
+ *
+ * Usage:
+ *
+ *    import {executeServerCommand} from '@web/test-runner-commands';
+ *
+ *    const cdnData = {
+ *      "foo": {
+ *        "1.2.3": {
```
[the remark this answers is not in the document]
aomarks: Oops, good catch! Done.

src/test/fake-cdn-plugin.ts
aomarks: > Sanity checking my understanding of the logic here. Does this mean that a concrete version has been found in `packageData.versions` that isn't for example a semverRange?
>
> Then this is a redirect from a range to a concrete version? I.e. a range becomes resolved into something concrete such as 1.2.3. Then we redirect calling this endpoint again with the new version. Now the version 1.2.3 would match the semverRange 1.2.3 and we'd reach line 89 where we fetch the content?

Yes exactly! And we actually rely on this redirect behavior in the `CachingCdn` class, because we inspect the redirected URLs in order to generate a canonical module specifier for each import.

Added a comment to explain this too!

PR #170, opened 2021-06-24
Add util module with path, specifier, etc. utility functions

aomarks:
Adds some new utility functions, and moves some existing ones, into a new `util.ts` module, and adds unit tests for all of them.

These are mostly related to URL paths and module specifiers, but there's also `MergedAsyncIterables` and `charToLineAndChar`. This and some of the other currently unused ones will be used in upcoming PRs that are part of https://github.com/PolymerLabs/playground-elements/issues/104.

Note each function + test is in its own commit,


src/typescript-worker/util.ts
```diff
@@ -118,3 +118,15 @@ export const parseNpmStyleSpecifier = (
   ];
   return {pkg, version: version ?? '', path};
 };
+
+/**
+ * Return the file extension of the given URL path. Does not include the leading
+ * ".".
+ */
+export const fileExtension = (path: string): string => {
```
AndrewJakubowicz, line 126: Not sure if this is out of scope for this but does this need to handle declaration extension, i.e. `test.d.ts` returns `.ts` currently?

src/test/util_test.ts
```diff
@@ -327,3 +328,29 @@ suite('fileExtension', () => {
...
   }
 });
+
+suite('changeFileExtension', () => {
+  const cases: Array<{
+    path: string;
+    newExt: string;
+    expected: string;
+  }> = [
+    {
```
AndrewJakubowicz, line 339: Nit: Would we ever have `foo.d.ts`, `ts` -> `foo.ts`?

src/typescript-worker/util.ts
```diff
@@ -130,3 +130,15 @@ export const fileExtension = (path: string): string => {
     ? ''
     : path.slice(lastDotIdx + 1);
 };
+
+/**
+ * Change the given URL path's file extension to a different one. `newExt`
+ * should not include the leading ".".
+ */
+export const changeFileExtension = (path: string, newExt: string): string => {
```
AndrewJakubowicz, line 138: I am not certain of the visibility and thus defensiveness required for this function, but should we throw on empty strings?

`changeFileExtension('test', '') // "test."`

`changeFileExtension('', 'ts') // ".ts"`

src/typescript-worker/util.ts
```diff
@@ -142,3 +142,24 @@ export const changeFileExtension = (path: string, newExt: string): string => {
   }
   return path.slice(0, -oldExt.length) + newExt;
 };
+
+/**
+ * Given a string and string-relative character index, return the equivalent
+ * line number and line-relative character index.
+ */
+export const charToLineAndChar = (
```
AndrewJakubowicz, line 150: Would there ever be a case where the newline is`\r\n` and does it matter?

AndrewJakubowicz (approved):
Exceptional! All my comments are for my own understanding or clarifications around certain edge cases that may not be relevant.

src/typescript-worker/util.ts
aomarks: The visibility is just our worker, and I think we don't need to throw because we're only going to be calling this for `.js` files, and the behavior we have is kinda reasonable, even though it would be a weird input.

src/typescript-worker/util.ts
aomarks: Good point, it could probably happen, but I think this implementation should still work fine*, since we'd consider the `\r` to be a regular character, and the `\n` would still determine the line boundaries. Added some test cases.

*Unless we want to handle commodore 64? https://en.wikipedia.org/wiki/Newline#Representation

aomarks (commented):
Thanks for the great review!

PR #171, opened 2021-06-24
Ignore query params when serving playground files

aomarks:
Previously we were matching filenames using the pathname + query params, due to the way we parsed URLs in the service worker, so we would serve 404s if there were query params. Now we strip query params.


e111077 (approved):
O WAOW

PR #174, opened 2021-07-14
Add CachingCdn class

aomarks:
More progress towards improved playground dependency handling.

This PR adds a `CachingCdn` class, which is an interface to unpkg.com or similar CDNs. The main interesting thing it helps with is specifier canonicalization -- e.g. for resolving `foo@^1.0.0` to `foo@1.2.3/index.js`, and it is intelligent about caching version ranges that have been already canonicalized (even for different paths).

No dedicated tests for this, because it's already being tested as part of some broader integration tests in an upcoming PR.


src/internal/build.ts
```diff
@@ -154,7 +154,7 @@ export class PlaygroundBuild {
 
   private _errorPendingFileRequests(error: HttpError) {
     for (const file of this._files.values()) {
-      if (!file.resolved) {
+      if (!file.settled) {
```
AndrewJakubowicz, line 157: No action required - just a question: If a file has already settled, does that mean it succeeded?

src/typescript-worker/util.ts
```diff
@@ -91,19 +91,25 @@ export const classifySpecifier = (
...
+}
+
 /**
  * Parse the given module import specifier using format
  * "<pkg>[@<version>][/<path>]".
  *
  * E.g. given "foo@^1.2.3/bar.js" return {
  *   pkg: "foo",
- *   range: "^1.2.3",
+ *   version: "^1.2.3",
```
AndrewJakubowicz, line 106: Great docs fix! This had me confused as I was reading.

AndrewJakubowicz (approved):
This is fantastic. I initially had some questions regarding the use of Deferred, but it looks like it's a really useful way to cache the fetch requests. This became clearer when looking over how the `_fetch` method is masterfully reused, and the cache is always checked.

I read through the two commits and my final verdict is that the code is outstanding. I am excited to read over the other commits and the integration tests.

src/internal/build.ts
aomarks: In this case, yeah! Either the worker produces a file by this name and resolves with the file, or the worker completes the build but never produces one of the files that the iframe requested, in which case we resolve it with this 404 error object instead.

PR #178, opened 2021-07-21
Force service worker updates on version mismatch

aomarks:
Previously, it was possible for the versions of a playground project element and its running service worker to be out of sync. This is because service workers are long-running, and only update themselves opportunistically. This could cause playgrounds to not work, or to behave strangely, for some time after an update.

We now include a version number check in the handshake between the playground project and the service worker. If the versions don't match, then we call [ServiceWorkerRegistration.update](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/update) to force an immediate update, and we wait for that update before we try and interact further with the service worker.

Also updates the way we deploy https://polymerlabs.github.io/playground-elements/ so that it's using the service worker from github, instead of the latest one published to unpkg (which is the default when `sandbox-base-url` is not specified).

Fixes https://github.com/PolymerLabs/playground-elements/issues/116


.gitignore
```diff
@@ -4,6 +4,7 @@
 /service-worker/
 /shared/
 /configurator/*.js
+!/configurator/playground-service-worker.js
```
AndrewJakubowicz, line 7: For my own understanding, what is configurator and how is it used when the `sandbox-base-url` is empty?

aomarks: The configurator is the site running at https://polymerlabs.github.io/playground-elements/

The `sandbox-base-url` parameter is explained best at https://github.com/PolymerLabs/playground-elements#sandbox-security

src/playground-project.ts
```diff
@@ -427,36 +418,67 @@ export class PlaygroundProject extends LitElement {
...
+            proxy({
+              getFile: (name: string) => this._getFile(name),
+            }),
+            this._sessionId
+          );
+        } else {
+          // Version mismatch. Request the service worker be updated
+          // immediately. We'll get back here again after it updates via a
+          // CONNECT_PROJECT_TO_SW message from the proxy.
+          console.info(
```
AndrewJakubowicz, line 448: Great comment! And is this `console.info` good to merge or for debugging? It looks great either way!

src/playground-project.ts
```diff
@@ -427,36 +418,67 @@ export class PlaygroundProject extends LitElement {
...
+            type: UPDATE_SERVICE_WORKER,
+          });
+        }
       }
     };
     port.addEventListener('message', onMessage);
     port.start();
   }
 
+  private _postMessageToServiceWorkerProxy(
```
justinfagnani, line 461: nit: I didn't know what "proxy" referred to until I read the comments here. Could the name be reconciled with the `_iframe` field? Name the iframe `_proxyIFrame`, `_swProxy` or the method `_postMessageToServiceWorkerIframe`?

src/playground-service-worker-proxy.ts
```diff
@@ -46,6 +51,19 @@ import {
...
 
+  window.addEventListener(
+    'message',
+    (event: MessageEvent<PlaygroundMessage>) => {
+      if (event.data.type === UPDATE_SERVICE_WORKER) {
+        // When the project handshakes with the service worker, it may notice a
+        // version mismatch, in which case it will send this message to request
+        // the service worker update. Note that service workers eventually
+        // update automatically, but not necessarily right away.
+        registration.update();
```
justinfagnani, line 62: Is there any result to wait for here? Should the main window elements show some affordance that the SW is updating, or an error if it fails to?

justinfagnani (approved):

src/playground-service-worker-proxy.ts
AndrewJakubowicz, line 62: I'm not able to follow where/when this update causes `CONNECT_PROJECT_TO_SW` to be called again, resulting in [this check](https://github.com/PolymerLabs/playground-elements/pull/178/files#diff-8f05853e3ce4da716aed3677ad38b34cf6423aaf403286a5a8ae458b97a85f5cR436) running again with same version.

My guess from my limited understanding is that this update triggers a `message` event here: https://github.com/PolymerLabs/playground-elements/pull/178/files#diff-8f05853e3ce4da716aed3677ad38b34cf6423aaf403286a5a8ae458b97a85f5cR457 .

AndrewJakubowicz (commented):
This looks great to me! I have some clarifying questions for my own understanding left in the changes.

Has this been tested manually (or some other way), and would you like me to run the code as well?
Thank you!

src/playground-project.ts
aomarks: Done.

src/playground-service-worker-proxy.ts
aomarks: > Is there any result to wait for here? Should the main window elements show some affordance that the SW is updating, or an error if it fails to?

We should probably show an error if we don't eventually get a new SW after calling this. However, we don't currently have a place to show *any* failure states, of which there are are a bunch. Seems like something to address in the future -- maybe we'll want an error notification element or something. Added a TODO for now.

aomarks: Calling update forces the update lifecycle to begin immediately, as opposed to at some indeterminate/opportunistic time in the future. That lifecycle is described here: https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API#download_install_and_activate

TL;DR The service worker script will be fetched, and byte-compared to the current one. If it's different, then the service worker gets an `activate` + `install` event, and then our registration object here will get an `updatefound` event, and then we call our `connect` function here to do a new handshake with the project element.

Expanded the comment here.

aomarks (commented):
> Has this been tested manually (or some other way), and would you like me to run the code as well?

I've tested it manually. You're welcome to try it out, but I don't feel it's necessary.

PR #179, opened 2021-07-23
Set Origin-Agent-Cluster header in service worker

aomarks:
Chrome (and maybe only Chrome currently) is able to allocate a separate process or thread to iframes, but only if the iframe and the parent are on separate *sites*, where a site is defined by protocol + top-level domain (note this does *not* include subdomains).

This is the main root cause of https://github.com/lit/lit.dev/issues/411. I had mistakenly assumed that since `lit.dev` and `playground.lit.dev` are separate origins, Chrome would execute Playground preview iframes in separate processes. But in fact it cannot, because those origins are same-site.

However, there is a new header called `Origin-Agent-Cluster`, which allows a server to switch the key for process isolation from site to origin. See https://web.dev/origin-agent-cluster/

This PR adds that header to service worker responses, and documents this issue for people who choose to change the default sandbox origin from unpkg to their own.

For playground.lit.dev, we'll also need to update the server itself to send this header, since it's actually the *first* response from an origin that determines this setting (I'm not actually sure yet exactly how long it persists for). Done in https://github.com/lit/lit.dev/pull/418.

See also:
https://www.chromium.org/developers/design-documents/oop-iframes
https://html.spec.whatwg.org/multipage/origin.html#origin-keyed-agent-clusters

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,11 @@ meets all of the following requirements:
    modifying the parent window using `window.parent`, e.g. to change your
    sign-in link to a malicious URL.
 
+   > NOTE: It is highly recommended to furthermore use either an entirely
+   > different _site_, or to use the `Origin-Agent-Cluster` header, to improve
+   > performance and prevent lockups. See [Process
+   > isolation](#process-isolation) for more information.
+
 2. Must not have access to any sensitive cookies. This prevents untrusted code
    from e.g. reading and forwarding your user's authentication token.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -580,6 +585,29 @@ meets all of the following requirements:
...
+or thread for iframes. This is highly desirable for Playground, because it
+improves responsiveness and prevents full lockups (resulting from e.g. an
+infinite loop accidentally written by a user).
+
+By default, this iframe process isolation can only occur if the iframe and the
+parent window are _different sites_. While an _origin_ is defined by (protocol +
+subdomain + top-level domain + port), a _site_ is defined only by (protocol +
+top-level domain). For example, `example.com` and `foo.example.com` are
+different-origin but same-site, whereas `example.com` and `example.net` are
+different-origin and different-site.
```
AndrewJakubowicz, line 600: 👏 👏 Really great docs.

AndrewJakubowicz (approved):
Fantastic!
Left my questions on https://github.com/lit/lit.dev/pull/418

Exceptional documentation for this improvement in the readme and code.

PR #180, opened 2021-07-28
Misc styling and bug fixes

aomarks:
- Fix `--playground-tab-bar-font-size`, and allow `::part(tab-bar)` to set `font-family` and `font-size`. This was broken because it was relying on being inherited to the inner `button` element, but `buttons` have a `font-size` and `font-family` set by the user agent stylesheet, which needed to be reset for inheritance to work.

- `playground-base-url` is now resolved relative to the `import.meta.url` of the module containing the definition of  `<playground-project>`, instead of relative to `document.location.href` as before. This means that `.` can now be used to refer to the local playground installation directory. (This makes it much easier to do local development with lit.dev + playground at the same time).

- Fix missing `.js` extension from import in `build.js` that may have broken Webpack and other builds.

- Fix invalid `--playground-floating-controls-color syntax`. This was broken due to a `var(var(` typo.

- Make active tab indicator inset, so that the hover effect and touch area for tabs in the tab bar are no longer offset by the height of the active tab indicator

- Fix bug that caused failures to load `d.ts` files and "Could not resolve module specifier" errors in the console. This was caused by a recent typo between `relativeUrlPath` and `resolveUrlPath`.

- Add eslint and run in CI.


README.md
```diff
@@ -800,7 +800,7 @@ to quickly experiment with themes and other customizations.
...
 | `--playground-tab-bar-active-color`                | ![](images/colors/6200EE.png) `var(--playground-highlight-color, #6200EE)`                  | Text `color` of active file-picker tab                                                                                             |
 | `--playground-tab-bar-indicator-color`             | ![](images/colors/6200EE.png) `var(--playground-highlight-color, #6200EE)`                  | `color` of active file-picker tab indicator (use `transparent` to hide)                                                            |
-| `--playground-tab-bar-font-size`                   | `0.85em`                                                                                    | `font-size` of tab titles in the file-picker tab bar                                                                               |
+| `--playground-tab-bar-font-size`                   | `14px`                                                                                      | `font-size` of tab titles in the file-picker tab bar                                                                               |
```
AndrewJakubowicz, line 803: Nit: Why the change of font-size default?

README.md
```diff
@@ -634,15 +634,15 @@ All-in-one project, editor, file switcher, and preview with a horizontal side-by
...
+| `lineNumbers`        | `boolean`                        | `false`                   | Render a gutter with line numbers in the editor                                                                                                                                                    |
+| `editableFileSystem` | `boolean`                        | `false`                   | Allow adding, removing, and renaming files                                                                                                                                                         |
+| `resizable`          | `boolean`                        | `false`                   | Allow dragging the line between editor and preview to change relative sizes                                                                                                                        |
+| `sandboxBaseUrl`     | `string`                         | _module parent directory_ | Base URL for untrusted JavaScript execution (⚠️ use with caution, see [sandbox security](#sandbox-security)). Resolved relative to the module containing the definition of `<playground-project>`. |
```
AndrewJakubowicz, line 644: Clarifying my understanding. Is the advantage of this over `location.href` that `meta.url` is more unique and specific to the executed code location. Does this lead to fewer url collisions?

src/internal/build.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {Deferred} from '../shared/deferred';
+import {Deferred} from '../shared/deferred.js';
```
AndrewJakubowicz, line 7: Great fix! Is there a way to programmatically check for this in the future, i.e. lint pass or a test webpack build?

src/internal/tab.ts
```diff
@@ -64,6 +64,19 @@ export class PlaygroundInternalTab extends LitElement {
     button:active::before {
       opacity: 20%;
     }
+
+    :host([active]) > button::after {
```
AndrewJakubowicz, line 68: I didn't check this manually but it looks good. A screenshot may help me better understand this. Is `:host` implicitly create a new stacking context for the `position: absolute;`?
Thank you!

src/typescript-worker/types-fetcher.ts
```diff
@@ -183,9 +183,7 @@ export class TypesFetcher {
       // Unhandled kind of import.
       return;
     }
-    // Note the base URL doesn't matter here, we're just using the URL API to
-    // perform path resolution.
-    const jsPath = relativeUrlPath(referrerSpecifier.path, relative).slice(1); // Remove the leading '/'.
+    const jsPath = resolveUrlPath(referrerSpecifier.path, relative).slice(1); // Remove the leading '/'.
```
AndrewJakubowicz, line 186: Nice!

AndrewJakubowicz (approved):
Tiny questions! LGTM!

README.md
aomarks: I thought it was weird that we used `em` for this value, but `px` for the code font size, and most of the other properties. I think that default was inherited from `mwc-tab`, which we used to use, before we switched to our own tab implementation. Updated the CHANGELOG to mention this.

src/internal/build.ts
aomarks: Good idea. Set up eslint with https://github.com/import-js/eslint-plugin-import/blob/master/docs/rules/extensions.md. Fixed all the other issues eslint found too.

src/internal/tab.ts
aomarks: Before:

image

After:

image

This is the state when the index.html tab is hovered. Note the white gap under it. (font-family also changed because of the other fix).

The `::after` pseudo-element does get a new stacking context since it's `absolute`, yeah. `<mwc-tab>` (which we used before https://github.com/PolymerLabs/playground-elements/pull/159) actually uses an absolutely positioned element for the indicator.

README.md
aomarks: The `sandboxBaseUrl` property tells playground where to look for the service worker script. By default, it's set to an unpkg.com URL, because of security (separate origin). However, it's useful during development to instead point at the local version, so that we can test changes to the service worker. This local version can be found in the root of wherever the `playground-elements` NPM package was installed (e.g. `./node_modules/playground-elements/playground-service-worker.js`) -- so that's what we want to set `sandboxBaseUrl` to. But it can be a little tricky to know exactly what that path would be. However, since we know the service worker and `playground-project.js` will be in the same place, we can use `import.meta.url` from `playground-project.js` to find that place. So changing the base URL to use `import.meta.url` instead of `document.location.href`, we can now pass `.` as the value of `sandboxBaseUrl` as a very easy way to say "use the service worker that's in the same directory you are".

PR #186, opened 2021-08-03
Dependency handling refactor

aomarks:
A major refactor to the way external dependencies are handled in Playground.

**Note to reviewer**: I recommend looking at each commit separately.

## Background

Before this PR, Playground directly transforms bare module specifiers to `https://unpkg.com/...?module` URLs. This works well for simple cases, but quickly breaks down because of issues around *module duplication*.

When the browser imports a module with some specifier, it caches that module *using exactly the given specifier as the key*. It doesn't matter if the URL ends up redirecting elsewhere -- that *original* specifier remains the key that determines the identity of the module. So, if another import statement imports what is ultimately the same module, but using a specifier that is different in any way, the browser will end up loading the same module *twice*. This is particularly problematic for custom element libraries, since registering an element with the same name twice throws an exception.

The following specifiers all represent the same underlying module, and will ultimately redirect to the same URL. But the browser will treat all of these specifiers as separate modules. *All of this variation can and does currently occur when using playground*, because import specifiers are not canonicalized -- neither in project files by us, nor in dependencies by unpkg:

```ts
/// What we transform `import 'foo';` to:
import 'https://unpkg.com/foo?module';

// What is ultimately redirected to:
import 'https://unpkg.com/foo@1.2.3/index.js?module';

// What another dependency's import of the same module might look like:
import 'https://unpkg.com/foo@^1.0.0?module';

// ... or this (a different, but compatible semver range):
import 'https://unpkg.com/foo@^1.2.0?module';

// ... or this (the default module is explicit, instead of implicit):
import 'https://unpkg.com/foo@^1.0.0/index.js?module';
```

### Real example 1: importing `<paper-input>`

[This](https://lit.dev/playground/#project=W3sibmFtZSI6ImluZGV4LnRzIiwiY29udGVudCI6ImltcG9ydCAnQHBvbHltZXIvcGFwZXItaW5wdXQvcGFwZXItaW5wdXQuanMnOyJ9LHsibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPHNjcmlwdCB0eXBlPVwibW9kdWxlXCIgc3JjPVwiLi9pbmRleC5qc1wiPjwvc2NyaXB0PlxuPHBhcGVyLWlucHV0IGxhYmVsPVwiRm9vXCI-PC9wYXBlci1pbnB1dD4ifV0) simple example of importing and using `<paper-input>` currently fails because:

1. `https://unpkg.com/@polymer/paper-input@3.2.1/paper-input.js?module` contains the import specifier `https://unpkg.com/@polymer/polymer@^3.0.0/lib/elements/dom-module.js?module`. Note the `@^3.0.0` in the specifier: unpkg simply substitutes the semver range from the `package.json` directly into the specifier, instead of resolving it to a concrete version first.
2. Via another import, `https://unpkg.com/@polymer/paper-input@3.2.1/paper-input.js?module`  imports `https://unpkg.com/@polymer/polymer@3.4.1/lib/mixins/element-mixin.js?module`.
3. `https://unpkg.com/@polymer/polymer@3.4.1/lib/mixins/element-mixin.js?module` imports `../elements/dom-module.js?module` aka `https://unpkg.com/@polymer/polymer@3.4.1/lib/elements/dom-module.js?module`.
4. An `Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "dom-module" has already been used with this registry` exception is thrown, because `dom-module.js` was loaded twice, using two different specifiers, which resulted in a double registration of the `<dom-module>` element.

### Real example 2: localization global state

The localization example: https://lit.dev/docs/libraries/localization/#example (filed at https://github.com/lit/lit.dev/issues/395) currently fails because:

1. The project file imports `@lit/localize`, which is transformed to `https://unpkg.com/@lit/localize`, and redirected to `https://unpkg.com/@lit/localize@0.10.3/lit-localize.js`.
2. `https://unpkg.com/@lit/localize@0.10.3/lit-localize.js` imports `./init/runtime.js`
3. `https://unpkg.com/@lit/localize@0.10.3/init/runtime.js` imports `../lit-localize.js` (yes, a circular import -- but that's valid in JS modules), and calls a function that modifies some module-scoped state in `lit-localize.js`
4. The module whose state was modified in 3 *is a different module* versus the one that we imported in 1 according to the browser, because while the two redirect to the same URL, their initial import specifier was different. This causes the example to fail.

## Changes

1. Module specifiers are now transformed to a path under `<service-worker-scope>/node_modules/`, instead of directly to unpkg URLs. This way, all dependency modules fall under the purview of our service worker, allowing us to provide those files ourselves.

2. Dependency modules are now parsed and transformed by us. We no longer use the unpkg `?module` mode. This allows us to implement more thorough canonicalization, and also lets us apply import maps to dependencies. Fixes https://github.com/PolymerLabs/playground-elements/issues/104.

3. Module specifiers are now always fully canonicalized. Versions are always resolved to their latest, concrete version (e.g. `^1.0.0` and `^1.1.0` are both canonicalized to `1.2.3`), and the default module is always made explicit (e.g. `foo` is canonicalized to `foo/index.js`). This is true both for imports in project files, and transitively within the entire dependency graph. Fixes https://github.com/PolymerLabs/playground-elements/issues/182.

4. We now check for a `package.json` in the project files, and use it for version constraints, just like how local NPM works. Previously, changing versions could only be done with an import map. The new approach should be more intuitive. It also means that when you click the "Download" button to get a tarball when a `package.json` is present, you'll get a configuration with correct versions locally too! Fixes https://github.com/PolymerLabs/playground-elements/issues/183.

## TODO

There will be a couple of follow-up PRs after this one:

1. Updating the `TypesFetcher` to follow the same logic. Currently, TypeScript type files are still fetched using the old approach, so the two are not in sync. I'll make sure to send this PR before the next release. (https://github.com/PolymerLabs/playground-elements/issues/184)

2. Add some caching. Since we now have to parse our dependency modules ourselves, there's a bit of a slowdown. However, a dependency at a given version never changes, so we can add some aggressive caching, meaning each dependency only needs to be parsed once per playground session + version. (https://github.com/PolymerLabs/playground-elements/issues/185)

### DIFF
--- a/src/test/bare-module-worker_test.ts
+++ b/src/test/bare-module-worker_test.ts
@@ -35,7 +35,7 @@ suite('bare module worker', () => {
     await checkTransform(files, expected, {}, cdn);
   });
 
-  test('one simple import', async () => {
+  test('one simple static import', async () => {
     const files: SampleFile[] = [
       {
         name: 'index.js',
@@ -75,14 +75,15 @@ suite('bare module worker', () => {
     await checkTransform(files, expected, {}, cdn);
   });
 
-  test('multiple simple imports', async () => {
+  test('multiple static and dynamic imports', async () => {
     const files: SampleFile[] = [
       {
         name: 'index.js',
         content: `
            import {a} from "foo/index.js";
-           import {b} from "foo/index.js";
+           const {b} = import("foo/index.js");
            import {c} from "foo/index.js";
+           const {d} = import("foo/index.js");
          `,
       },
     ];
@@ -96,6 +97,7 @@ suite('bare module worker', () => {
                    export const a = 'a';
                    export const b = 'b';
                    export const c = 'c';
+                   export const d = 'd';
                    `,
               },
             },
@@ -110,8 +112,9 @@ suite('bare module worker', () => {
           name: 'index.js',
           content: `
            import {a} from "./node_modules/foo@1.0.0/index.js";
-           import {b} from "./node_modules/foo@1.0.0/index.js";
+           const {b} = import('./node_modules/foo@1.0.0/index.js');
            import {c} from "./node_modules/foo@1.0.0/index.js";
+           const {d} = import('./node_modules/foo@1.0.0/index.js');
          `,
         },
       },
@@ -123,6 +126,7 @@ suite('bare module worker', () => {
                    export const a = 'a';
                    export const b = 'b';
                    export const c = 'c';
+                   export const d = 'd';
                    `,
           contentType: 'text/javascript; charset=utf-8',
         },
@@ -171,7 +175,7 @@ suite('bare module worker', () => {
     await checkTransform(files, expected, {}, cdn);
   });
 
-  test('"module" field is 1st choice for default module', async () => {
+  test('default module uses package.json "module" field as 1st choice', async () => {
     const files: SampleFile[] = [
       {
         name: 'index.js',
@@ -223,7 +227,7 @@ suite('bare module worker', () => {
     await checkTransform(files, expected, {}, cdn);
   });
 
-  test('"main" field is 2nd choice for default module', async () => {
+  test('default module uses package.json "main" field as 2nd choice', async () => {
     const files: SampleFile[] = [
       {
         name: 'index.js',
@@ -274,7 +278,7 @@ suite('bare module worker', () => {
     await checkTransform(files, expected, {}, cdn);
   });
 
-  test('"index.js" is 3rd choice for default module', async () => {
+  test('default module uses "index.js" field as 3rd choice', async () => {
     const files: SampleFile[] = [
       {
         name: 'index.js',
--- a/src/test/bare-module-worker_test.ts
+++ b/src/test/bare-module-worker_test.ts
@@ -4,67 +4,103 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {checkTransform} from './worker-test-util.js';
+
 import type {
   BuildOutput,
   ModuleImportMap,
   SampleFile,
 } from '../shared/worker-api.js';
+import type {CdnData} from './fake-cdn-plugin.js';
 
-import {checkTransform} from './worker-test-util.js';
+const browser = (() => {
+  const ua = navigator.userAgent;
+  if (ua.includes('Chrome')) {
+    return 'chrome';
+  }
+  if (ua.includes('Firefox')) {
+    return 'firefox';
+  }
+  if (ua.includes('Safari')) {
+    return 'safari';
+  }
+  return 'chrome';
+})();
 
-suite('bare module builder', () => {
-  test('transforms static bare module import to unpkg', async () => {
+suite('bare module worker', () => {
+  test('empty project', async () => {
+    const files: SampleFile[] = [];
+    const cdn: CdnData = {};
+    const expected: BuildOutput[] = [];
+    await checkTransform(files, expected, {}, cdn);
+  });
+
+  test('one simple import', async () => {
     const files: SampleFile[] = [
       {
         name: 'index.js',
-        content: 'import "lit-html";',
+        content: 'import "foo/index.js";',
       },
     ];
+    const cdn: CdnData = {
+      foo: {
+        versions: {
+          '1.0.0': {
+            files: {
+              'index.js': {
+                content: 'index',
+              },
+            },
+          },
+        },
+      },
+    };
     const expected: BuildOutput[] = [
       {
         kind: 'file',
         file: {
           name: 'index.js',
-          content: 'import "https://unpkg.com/lit-html?module";',
+          content: 'import "./node_modules/foo@1.0.0/index.js";',
         },
       },
-    ];
-    await checkTransform(files, expected);
-  });
-
-  test('transforms dynamic bare module import to unpkg', async () => {
-    const files: SampleFile[] = [
-      {
-        name: 'index.js',
-        content: 'import("lit-html");',
-      },
-    ];
-    const expected: BuildOutput[] = [
       {
         kind: 'file',
         file: {
-          name: 'index.js',
-          content: "import('https://unpkg.com/lit-html?module');",
+          name: 'node_modules/foo@1.0.0/index.js',
+          content: 'index',
+          contentType: 'text/javascript; charset=utf-8',
         },
       },
     ];
-    await checkTransform(files, expected);
+    await checkTransform(files, expected, {}, cdn);
   });
 
-  test('follows import map', async () => {
+  test('multiple simple imports', async () => {
     const files: SampleFile[] = [
       {
         name: 'index.js',
         content: `
-          import "lit-html";
-          import "lit-html/directives/repeat.js";
-        `,
+           import {a} from "foo/index.js";
+           import {b} from "foo/index.js";
+           import {c} from "foo/index.js";
+         `,
       },
     ];
-    const importMap: ModuleImportMap = {
-      imports: {
-        'lit-html': 'https://cdn.skypack.dev/lit-html@2.0.0-pre.7',
-        'lit-html/': 'https://cdn.skypack.dev/lit-html@2.0.0-pre.7/',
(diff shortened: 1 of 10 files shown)


src/test/bare-module-worker_test.ts
```diff
@@ -4,119 +4,872 @@
...
+          name: 'node_modules/foo@1.0.0/main.js',
+          content: 'main',
+          contentType: 'text/javascript; charset=utf-8',
+        },
+      },
+    ];
+    await checkTransform(files, expected, {}, cdn);
+  });
+
+  test('"index.js" is 3rd choice for default module', async () => {
```
AndrewJakubowicz, line 277: Nit: Clarification of whether this number refers to the `CdnData` file key map order or the BuildOutput array?

My expectation comparing this test to the previous one is that this refers to the `files` object within CdnData.

src/test/bare-module-worker_test.ts
```diff
@@ -4,119 +4,872 @@
...
+          },
+        },
+      },
+    };
+    const expected: BuildOutput[] = [
+      {
+        kind: 'file',
+        file: {
+          name: 'some/sub/dir/index.js',
+          content: 'import "../../../node_modules/foo@1.0.0/index.js";',
```
AndrewJakubowicz, line 348: This is a fantastic test. It's now clicked for me how the imports are being normalized.

src/test/bare-module-worker_test.ts
```diff
@@ -4,119 +4,872 @@
...
+        kind: 'file',
+        file: {
+          name: 'index.js',
+          content: 'import "non-existent/index.js";',
+        },
+      },
       {
         diagnostic: {
-          message: 'es-module-lexer error: Parse error @:1:7',
+          message:
```
AndrewJakubowicz, line 742: Wow this is an improved error message.

src/test/bare-module-worker_test.ts
```diff
@@ -4,119 +4,872 @@
...
+        },
+      },
+    };
+    const expected: BuildOutput[] = [
       {
         kind: 'file',
         file: {
           name: 'index.js',
-          content: 'import"',
+          content: 'import "http://example.com/foo";',
```
AndrewJakubowicz, line 869: Clarification - is there anything else we do with dynamic imports or just swap out the import string?

Could this be a relative url `/node_modules/foo.js`?

src/typescript-worker/bare-module-transformer.ts
```diff
@@ -5,35 +5,115 @@
...
-    this._moduleResolver = moduleResolver;
+  constructor(cdn: CachingCdn, importMapResolver: ModuleResolver) {
+    this._cdn = cdn;
+    this._importMapResolver = importMapResolver;
   }
 
   async *process(
     results: AsyncIterable<BuildOutput> | Iterable<BuildOutput>
   ): AsyncIterable<BuildOutput> {
+    // We're going to asynchronously walk the entire dependency graph. This
```
AndrewJakubowicz, line 65: nit: Everything is great. Does it make sense to make this a jsdoc process method comment?

src/typescript-worker/bare-module-transformer.ts
```diff
@@ -42,63 +122,293 @@ export class BareModuleTransformer {
...
+    }
+    // E.g. if `referrer` is "node_modules/foo@1.2.3/a.js" and `specifier` is
+    // "./b.js", then `absolute` will be "node_modules/foo@1.2.3/b.js", and
+    // `bare` will be "foo@1.2.3/b.js".
+    const absolute = resolveUrlPath(referrer, specifier);
+    const bare = absolute.slice('/node_modules/'.length);
+    if (!fileExtension(specifier)) {
+      // We can't simply return the existing relative specifier if there's no
+      // ".js" extension, because we still need to do path canonicalization. For
+      // example: "./foo" could refer to "./foo.js" or "./foo/index.js"
```
AndrewJakubowicz, line 229: Great comment!

src/typescript-worker/bare-module-transformer.ts
```diff
@@ -42,63 +122,293 @@ export class BareModuleTransformer {
...
+  json: string
+): {line: number; character: number} | undefined => {
+  const chrome = message.match(/at position (\d+)/);
+  if (chrome !== null) {
+    return charToLineAndChar(json, Number(chrome[1]));
+  }
+  const firefox = message.match(/at line (\d+) column (\d+)/);
+  if (firefox !== null) {
+    return {
+      line: Number(firefox[1]) - 1,
```
AndrewJakubowicz, line 409: Is there a difference between `Number(var)` and `parseInt(var, 10)`?

src/typescript-worker/build.ts
```diff
@@ -22,9 +23,10 @@ export const build = async (
   },
   emit: (result: BuildOutput) => void
 ): Promise<void> => {
-  const moduleResolver = new ModuleResolver(importMap);
+  const moduleResolver = new ModuleResolver(config.importMap);
+  const cdn = new CachingCdn(config.cdnBaseUrl ?? 'https://unpkg.com/');
```
AndrewJakubowicz, line 27: In the readme it mentions that unpkg is used behind the scenes. Is the `cdnBaseUrl` configurable via a public interface such that another cdn could be used?

AndrewJakubowicz (approved):
Everything looks fantastic. Left a sprinkling of questions and clarifications but this is really amazing.

Is the `bare-module-transformer.ts` file unit tested, or does it use the output of the unit tested code?

Didn't run manually, spent my focus on how great this is.

src/test/bare-module-worker_test.ts
aomarks: > Clarification - is there anything else we do with dynamic imports or just swap out the import string?

Oh, I forgot to include any tests for dynamic import! Updated a test to include some (near the top of this file).

> Could this be a relative url `/node_modules/foo.js`?

This test is specifically checking the behavior of import maps, so we're not expecting the `node_modules` path at all here.

For the non-import-map cases though, we don't want use a root-relative URL like `/node_modules` (as opposed to `./node_modules`) because we don't know the root-relative path to the service-worker scope, since that is determined by the `sandboxBaseUrl` parameter and is customizable. E.g. by default we need to resolve to `https://unpkg.com/playground-elements@<version>/playground-projects/node_modules/...`, but on lit.dev it's `https://playground.lit.dev/node_modules/...`. We could pipe that parameter through to generate root-relative URLs, but it's simpler to compute paths that are file-relative.

src/typescript-worker/bare-module-transformer.ts
aomarks: This comment is here specifically to help explain why we have this `MergedAsyncIterables` object, so I think it does belong just above the variable declaration. I rephrased the comment to make it more clear what the primary subject is, and to be a little more useful.

src/typescript-worker/bare-module-transformer.ts
aomarks: Since we know we're only handling something that matches `\d+`, there's no difference in the output, and `Number` seemed shorter/simpler.

I like this table for seeing what all the differences are: https://stackoverflow.com/questions/17106681/parseint-vs-unary-plus-when-to-use-which/17106702#17106702

aomarks (commented):
> Is the `bare-module-transformer.ts` file unit tested, or does it use the output of the unit tested code?

It's tested via the build function so that it covers more surface area, since it's always invoked via the build function.

> Offline also asked if package.json `browser` field is handled.

I've been using Rollup as my rough guide to behavior on questions like this. It's node resolve plugin ignores `browser` by default, but you can turn it on with an option: https://github.com/rollup/plugins/tree/master/packages/node-resolve/#browser. Personally I haven't ever encountered a case where I needed the `browser` field to be followed. Easy to add in later if it comes up.

I'm even more interested in supporting the new `exports` field (https://nodejs.org/api/packages.html#packages_conditional_exports), filed https://github.com/PolymerLabs/playground-elements/issues/188.

PR #189, opened 2021-08-09
feat(sw): add more mimetypes for media

e111077:
particularly want to make svg work, but discussed offline that other media can exist and be hidden by author

### DIFF
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -740,8 +740,6 @@ const typeEnumToMimeType = (type?: string) => {
       return 'video/mpeg';
     case 'ogv':
       return 'video/ogg';
-    case 'ts':
-      return 'video/mp2t';
     case 'webm':
       return 'video/webm';
     case '3gp':
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -718,6 +718,8 @@ const typeEnumToMimeType = (type?: string) => {
       return 'image/svg+xml';
     case 'png':
       return 'image/png';
+    case 'gif':
+      return 'image/gif';
     case 'jpeg':
     case 'jpg':
       return 'image/jpeg';
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -730,8 +730,6 @@ const typeEnumToMimeType = (type?: string) => {
     case 'mid':
     case 'midi':
       return 'audio/midi';
-    case 'webp':
-      return 'audio/webp';
     case 'mp3':
       return 'audio/mpeg';
     case 'weba':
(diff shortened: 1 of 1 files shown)


src/playground-project.ts
```diff
@@ -712,6 +712,59 @@ const typeEnumToMimeType = (type?: string) => {
...
+      return 'image/webp';
+    case 'avi':
+      return 'video/x-msvideo';
+    case 'mp4':
+      return 'video/mp4';
+    case 'mpeg':
+      return 'video/mpeg';
+    case 'ogv':
+      return 'video/ogg';
+    case 'ts':
```
aomarks: We already have to map `ts` for typescript in a case above. We'll reach here, but let's not even include it (or maybe comment it out with a comment explaining that).

src/playground-project.ts
```diff
@@ -712,6 +712,59 @@ const typeEnumToMimeType = (type?: string) => {
...
+      return 'video/3gpp';
+    case '3g2':
+      return 'video/3gpp2';
+    case 'aac':
+      return 'audio/aac';
+    case 'mid':
+    case 'midi':
+      return 'audio/midi';
+    case 'webp':
+      return 'audio/webp';
```
aomarks, line 755: There's already a webp image case above.

src/playground-project.ts
```diff
@@ -712,6 +712,57 @@ const typeEnumToMimeType = (type?: string) => {
       return 'text/html; charset=utf-8';
     case 'css':
       return 'text/css; charset=utf-8';
+    // taken from MDN's common MIME types
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+    case 'svg':
```
aomarks, line 717: Just in the interest of brevity, I think I'd probably skip these:

```
bmp
tif
tiff
avi
mp4
mpeg
ogv
3gp
3g2
aac
oga
opus
wav
mid
midi
```

Because we have better alternatives (png/webp for images, weba for audio, webm for video). We can always tell users to use one of those, or to just explicitly specify the mime type in the config file if they really want (using the `contentType` field).

src/playground-project.ts
```diff
@@ -712,6 +712,59 @@ const typeEnumToMimeType = (type?: string) => {
       return 'text/html; charset=utf-8';
     case 'css':
       return 'text/css; charset=utf-8';
+    // taken from MDN's common MIME types
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
```
aomarks, line 716: In the interest of brevity, I think I'd skip these:

```
bmp
gif
tif
tiff
avi
mp4
mpeg
ogv
3gp
3g2
aac
mid
midi
oga
opus
wav
```

I think most of these media formats are already pretty edge case anyway, Let's just include the "modern" and old-but-too-common-on-the-web-to-ignore formats. Users can always explicitly set the mime type with the `contentType` field in the JSON config.

aomarks (changes requested):

aomarks (approved):
hooray

PR #193, opened 2021-08-16
Improved types fetcher

aomarks:
A big refactor of the `TypesFetcher` class.

- We now correctly check types when there are multiple versions of the same package simultaneously. Previously, if two versions of the same package appeared in the transitive import graph of a project, only some arbitrary version would be included. That was because we presented a completely flat virtual `node_modules/` directory to the TypeScript compiler. We now support nested `node_modules/` directories when there are version conflicts, much like a local NPM install. Fixes https://github.com/PolymerLabs/playground-elements/issues/192

- We now respect the project's `package.json` if it exists, to determine which version of a package an import statement should be for. Previously, we ignored the `package.json`, so only import maps could be used to change the version from `latest`. Fixes https://github.com/PolymerLabs/playground-elements/issues/184.

**Note to reviewer**: you may want to read `types-fetcher.ts` by itself, instead of as a diff, because it's quite significantly different.


src/typescript-worker/types-fetcher.ts
```diff
@@ -9,244 +9,385 @@ import {ModuleResolver} from './module-resolver.js';
...
+   * Fetch all ".d.ts" typing files for the full transitive dependency tree of
+   * the given project source files and standard libs.
    *
-   * @returns Promise of a Map whose keys are bare module specifiers, and values
-   * are file contents. Example keys: "lit-html/lit-html.d.ts",
-   * "lit-html/package.json".
+   * @param sources Project TypeScript source file contents. All bare module
+   * imports in these sources will be followed.
+   * @param tsLibs TypeScript standard libraries to fetch, e.g. "es2020", "DOM".
+   * Case-insensitive.
```
AndrewJakubowicz, line 71: Nit: Is case insensitivity tested or benefit from testing?

src/typescript-worker/types-fetcher.ts
```diff
@@ -9,244 +9,385 @@ import {ModuleResolver} from './module-resolver.js';
...
+    if (location === undefined) {
       return;
     }
-    this._handledSpecifiers.add(bare);
-    const npm = parseNpmStyleSpecifier(bare);
-    if (npm === undefined) {
-      return;
+    // Versions don't make much sense with an import map, since you just have
+    // bare specifier -> URL. We can leave whatever version we already have; it
+    // will always be ignored.
```
AndrewJakubowicz, line 155: If there are two different imports in the import map that reference the same url will the work be duplicated and aliased under the two different importMap keys?

I.e.:
```
"imports": {
    "a": "<URL_TO_A>",
    "b": "<URL_TO_A>",
}
```

src/typescript-worker/types-fetcher.ts
```diff
@@ -9,244 +9,385 @@ import {ModuleResolver} from './module-resolver.js';
...
-    } else if (ext !== 'js') {
-      // Unhandled kind of import.
+    const location = {
+      // We know package and version must be the same as the referrer, since
+      // this is a relative path, and we are therefore still in the same
+      // package.
+      pkg: referrer.pkg,
+      version: referrer.version,
+      // Make the path package-root relative, instead of referrer-path relative.
+      path: resolveUrlPath(referrer.path, relative).slice(1), // Remove the leading '/'.
```
AndrewJakubowicz, line 207: Nit: Rather than the comment is it worth using the `trimLeadingSlash` util.

src/typescript-worker/typescript.ts
```diff
@@ -109,12 +118,22 @@ export class TypeScriptBuilder {
 
     // Wait for all typings to be fetched, and then retrieve slower semantic
     // diagnostics.
-    for (const [specifier, content] of await typesFetcher.getFiles()) {
+    const typesFetcher = new TypesFetcher(
+      this._cdn,
+      this._importMapResolver,
+      packageJson
+    );
```
AndrewJakubowicz, line 125: Does the `TypesFetcher` do any caching of files already loaded?

src/test/worker-test-util.ts
```diff
@@ -15,7 +15,7 @@ import type {
 } from '../shared/worker-api.js';
 import type {CdnData} from './fake-cdn-plugin.js';
 
-const configureFakeCdn = async (
+export const configureFakeCdn = async (
```
AndrewJakubowicz, line 18: I read a comment somewhere mentioning how network errors are handled. Raised a small (non actionable) question.
Because we have a fake CDN we can possibly test or simulate various network errors. Do we test behavior under network failure. Very slow response or intermittent failure. I'm not sure how valuable that would be though.

src/typescript-worker/types-fetcher.ts
```diff
@@ -9,244 +9,385 @@ import {ModuleResolver} from './module-resolver.js';
...
-  /**
-   * Start fetching type definitions for a built-in TypeScript lib, like "dom"
-   * or "esnext".
-   *
-   * This function returns immediately, but begins an asynchronous walk of the
-   * <reference> graph. To access the results, await {@link getTypings}.
-   */
-  addLibTypings(lib: string): void {
-    this._entrypointTasks.push(this._addLibTypings(lib));
+  constructor(
```
AndrewJakubowicz, line 54: Non actionable thinking out loud / clarifying. This is a really great class and I especially love the thorough testing.

From reading over the class it seems to be a very stateful class, populated by first construction and then `fetchTypes`. This mutates all the relevant data such that retrieved files can be retrieved.

Are there any invariants that need to be kept in mind when calling the methods that could become subtle mutation errors? For example is it valid to call `fetchTypes` multiple times on the same instance? What if the first one has a 500 status code on one of the requests and on the second call it's a 200 status?

AndrewJakubowicz (approved):
This is phenomenal. It's been especially a joy reviewing as I love seeing the previously built and tested components combined into the `TypesFetcher`.

All my comments are small understanding comments and clarifications. I also didn't quite understand what kind of network failure conditions are handled, or how they surface to the end user.

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Import maps now apply to modules in external dependencies, not just to local
   project files.
 
+- The TypeScript compiler can now deal with multiple versions of the same
```
AndrewJakubowicz, line 32: What makes a change breaking? Could some subtle import or more nuanced behavior in the import resolution make this a breaking change?

src/typescript-worker/types-fetcher.ts
aomarks: This works -- the typings will appear both under `node_modules/a` and `node_modules/b`. I updated a test case to confirm.

As written, there is actually a duplicate fetch, because our fetch cache is currently keyed by specifier, not URL. However, I think I'm pretty ok with ignoring that because 1. Mapping two packages to the same URL seems really ununsual, and 2. The browser HTTP cache should still get a hit, just a little deeper in the stack.

src/typescript-worker/types-fetcher.ts
aomarks: Good call, done!

src/typescript-worker/typescript.ts
aomarks: We cache results by specifier (package + version + path) *within* each build to prevent duplicate fetches/analysis, but yeah -- there is currently no caching *across* builds. This is definitely something to address in a follow up. If the imports and package.json haven't changed since last build, then we can re-use all of this work, whereas currently we recompute all of it. Note that the underlying browser HTTP cache will take care of the main network cost of this duplicated work, but there's also a lot of parsing and analysis work being done that probably adds noticeable latency.

Filed https://github.com/PolymerLabs/playground-elements/issues/185.

src/typescript-worker/types-fetcher.ts
aomarks: Good point --  the current signature implies there's more re-use possible than there actually is, and could be misused. It's actually just a single-use class: one is created for every build, and the `fetchTypes` method is only ever called once.

I've refactored it a little so that the only public API is a static method that ensures `fetchTypes` is only called once. Hopefully this is more clear!

To answer the specific question (which is now impossible, but still interesting): If a request initially was a 500 error, then it would always have the same error code if `fetchTypes` was called multiple times, since we cache the HTTP response status code directly when it is not a 200.

aomarks (commented):
Thanks! I believe I have addressed all feedback.

> All my comments are small understanding comments and clarifications. I also didn't quite understand what kind of network failure conditions are handled, or how they surface to the end user.

Currently, if an error occurs directly on a top-level import, then the user will see a "Cannot find module" error as a red squiggly, because we won't have put the corresponding typings file in the virtual `node_modules/` folder, and TypeScript will complain about that. If it occurred somewhere deeper in the tree, then the user won't see anything currently.

It *would* be good to surface these deeper errors. I think the best thing to do would be to create a new component for showing logs/errors, and surface them there. That would mirror what would happen in a local dev environment, where this sort of error would occur either on npm install (e.g. the NPM server being down so that something couldn't be installed), or as an error in the TypeScript compile output (e.g. because some dependency imported a module that doesn't exist [unless `skipLibCheck` was on, in which case they'd be ignored there too]).

configurator/project/project.json
```diff
@@ -1,19 +1,7 @@
...
-      "@lit/reactive-element": "https://cdn.skypack.dev/@lit/reactive-element",
-      "@lit/reactive-element/": "https://cdn.skypack.dev/@lit/reactive-element/",
-      "lit-element": "https://cdn.skypack.dev/lit-element@3.0.0-pre.4",
-      "lit-element/": "https://cdn.skypack.dev/lit-element@3.0.0-pre.4/",
-      "lit-html": "https://cdn.skypack.dev/lit-html@2.0.0-pre.7",
-      "lit-html/": "https://cdn.skypack.dev/lit-html@2.0.0-pre.7/",
-      "typescript/": "https://unpkg.com/typescript@4.2.3/"
-    }
+    "index.html": {},
+    "package.json": {}
```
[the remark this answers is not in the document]
AndrewJakubowicz: Ahh. I got confused by the empty object literal. Thanks for the clarification!

src/typescript-worker/types-fetcher.ts
AndrewJakubowicz: Awesome! Also that is interesting!

src/test/types-fetcher_test.ts
```diff
@@ -573,6 +573,65 @@ suite('types fetcher', () => {
     });
   });
 
+  test('tslibs', async () => {
```
AndrewJakubowicz, line 574: 🎉

AndrewJakubowicz (approved):
Looked through changes and they look amazing!

PR #201, opened 2021-09-13
Add a modified property to project and IDE

aomarks:
Adds a `modified` property to `<playground-project>` and `<playground-ide>` which indicates whether the user has modified, added, or removed any project files. Resets whenever a new project is loaded.

We'll use this as part of fixing https://github.com/lit/lit.dev/issues/485.


src/playground-project.ts
```diff
@@ -313,6 +336,8 @@ export class PlaygroundProject extends LitElement {
         source as void; // Exhaustive check.
         break;
     }
+    this._pristineFiles =
+      this._files && JSON.parse(JSON.stringify(this._files));
```
AndrewJakubowicz, line 340: Is there a more memory efficient way to ensure a file changed?  Also it probably will never get so big to matter.

src/playground-project.ts
```diff
@@ -838,3 +863,29 @@ const outdent = (str: string): string => {
...
+  }
+  for (let i = 0; i < filesA.length; i++) {
+    const fileA = filesA[i];
+    const fileB = filesB[i];
+    if (
+      fileA.name !== fileB.name ||
+      fileA.content !== fileB.content ||
+      fileA.contentType !== fileB.contentType ||
+      fileA.hidden !== fileB.hidden ||
+      fileA.label !== fileB.label
```
AndrewJakubowicz, line 885: Nit: For greatest chance of short circuiting after minimal work, we could check the `content` last (which I'm assuming will have the most string comparing).

Could also do 2 loops. The first loop checks all the metadata that is relatively small ignoring any content checking. This gives a very quick way to find inequality without content. Then do a second loop over the content. 

Computers are pretty fast so no changes required.

AndrewJakubowicz (approved):
🎉 LGTM

aomarks (commented):
PTAL pushed some new caching logic.

AndrewJakubowicz (approved):
Excellent changes! Is there a way to test that the double modified checks are caching correctly and not recalculating twice?

New changes look solid! 🎉

aomarks:
> Hmm, I think not without exposing or monkey patching some internals that I would prefer not to do. I did check it manually with a log statement, though.

Clarify: the test does confirm that the result is *correct* by checking the value twice after each change -- it just doesn't confirm that it's doing it in the optimal way.

PR #205, opened 2021-09-18
Don't create a new languageservice on every build + incremental builds

Matsuuu:
Heya,

As per our discussion here and on slack, I've fiddled around with the code base a bit.

I did a simple map-based "cache" for the typescript builders, saving them by project id's and rewrote the way we handled the languageservices.

I made the now cached instance of TypescriptBuilder save the languageService and the languageServiceHost as class level properties, so that we can re-use them instead of instantiating a new one every build.

I also expanded the WorkerLanguageServiceHost class to handle the new way of adding, and syncing the project files. Now with the added versioning, we can just update the necessary files, and by bumping up the version number on those files, the languageservice will be able to incrementally build them.

A lot of inspiration was drawn from the [TS docs](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#incremental-build-support-using-the-language-services). I'd love to have a review on this and if this is going in the right direction?

From my local testing, everything should work as it did before, just a bit snappier and more efficiently.

If this change and coding style is fine, I'd love to delve deeper into the actual language service support and integrating more LSP features to the playground.

### DIFF
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -545,7 +545,6 @@ export class PlaygroundProject extends LitElement {
       return;
     }
     workerApi.compileProject(
-      this._sessionId,
       this._files ?? [],
       {importMap: this._importMap},
       proxy((result) => build.onOutput(result))
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -545,6 +545,7 @@ export class PlaygroundProject extends LitElement {
       return;
     }
     workerApi.compileProject(
+      this._sessionId,
       this._files ?? [],
       {importMap: this._importMap},
       proxy((result) => build.onOutput(result))
--- a/src/typescript-worker/build.ts
+++ b/src/typescript-worker/build.ts
@@ -16,7 +16,6 @@ import {CachingCdn} from './caching-cdn.js';
 import { getBuilder } from './project-cache.js';
 
 export const build = async (
-  projectId: string,
   files: Array<SampleFile>,
   config: {
     importMap: ModuleImportMap;
@@ -26,7 +25,7 @@ export const build = async (
 ): Promise<void> => {
   const moduleResolver = new ImportMapResolver(config.importMap);
   const cdn = new CachingCdn(config.cdnBaseUrl ?? 'https://unpkg.com/');
-  const tsBuilder = getBuilder(projectId, cdn, moduleResolver);
+  const tsBuilder = getBuilder(cdn, moduleResolver);
   const bareModuleBuilder = new BareModuleTransformer(cdn, moduleResolver);
   const results = bareModuleBuilder.process(
     tsBuilder.process(files.map((file) => ({kind: 'file', file})))
--- a/src/typescript-worker/build.ts
+++ b/src/typescript-worker/build.ts
@@ -9,23 +9,20 @@ import {ImportMapResolver} from './import-map-resolver.js';
 
 import type {
   SampleFile,
-  ModuleImportMap,
   BuildOutput,
+  TypeScriptWorkerConfig,
 } from '../shared/worker-api.js';
 import {CachingCdn} from './caching-cdn.js';
-import { getBuilder } from './project-cache.js';
+import {getBuilder} from './project-cache.js';
 
 export const build = async (
   files: Array<SampleFile>,
-  config: {
-    importMap: ModuleImportMap;
-    cdnBaseUrl?: string;
-  },
+  config: TypeScriptWorkerConfig,
   emit: (result: BuildOutput) => void
 ): Promise<void> => {
   const moduleResolver = new ImportMapResolver(config.importMap);
   const cdn = new CachingCdn(config.cdnBaseUrl ?? 'https://unpkg.com/');
-  const tsBuilder = getBuilder(cdn, moduleResolver);
+  const tsBuilder = getBuilder(config, cdn, moduleResolver);
   const bareModuleBuilder = new BareModuleTransformer(cdn, moduleResolver);
   const results = bareModuleBuilder.process(
     tsBuilder.process(files.map((file) => ({kind: 'file', file})))
--- a/src/typescript-worker/build.ts
+++ b/src/typescript-worker/build.ts
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {TypeScriptBuilder} from './typescript.js';
 import {BareModuleTransformer} from './bare-module-transformer.js';
 import {ImportMapResolver} from './import-map-resolver.js';
 
@@ -14,8 +13,10 @@ import type {
   BuildOutput,
 } from '../shared/worker-api.js';
 import {CachingCdn} from './caching-cdn.js';
+import { getBuilder } from './project-cache.js';
 
 export const build = async (
+  projectId: string,
   files: Array<SampleFile>,
   config: {
     importMap: ModuleImportMap;
@@ -23,9 +24,10 @@ export const build = async (
   },
   emit: (result: BuildOutput) => void
 ): Promise<void> => {
+    console.log({files, config, emit, projectId})
   const moduleResolver = new ImportMapResolver(config.importMap);
   const cdn = new CachingCdn(config.cdnBaseUrl ?? 'https://unpkg.com/');
-  const tsBuilder = new TypeScriptBuilder(cdn, moduleResolver);
+  const tsBuilder = getBuilder(projectId, cdn, moduleResolver);
   const bareModuleBuilder = new BareModuleTransformer(cdn, moduleResolver);
   const results = bareModuleBuilder.process(
     tsBuilder.process(files.map((file) => ({kind: 'file', file})))
--- a/src/typescript-worker/build.ts
+++ b/src/typescript-worker/build.ts
@@ -24,7 +24,6 @@ export const build = async (
   },
   emit: (result: BuildOutput) => void
 ): Promise<void> => {
-    console.log({files, config, emit, projectId})
   const moduleResolver = new ImportMapResolver(config.importMap);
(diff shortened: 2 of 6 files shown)


src/typescript-worker/project-cache.ts
```diff
@@ -0,0 +1,15 @@
+import { TypeScriptBuilder } from "./typescript";
+import {CachingCdn} from './caching-cdn.js';
+import { ImportMapResolver } from "./import-map-resolver";
+
+
+export const builderCache = new Map();
+
+export function getBuilder(projectId: string, cdn: CachingCdn, moduleResolver: ImportMapResolver) {
```
aomarks, line 8: Related to other comment -- since we know that there's a 1-1 relationship between each `<playground-project>` and its dedicated web worker, I think we don't actually need this cache. We can just make one module-scoped TypeScript builder :D

src/typescript-worker/typescript.ts
```diff
@@ -37,11 +38,27 @@ export class TypeScriptBuilder {
...
+
   constructor(cdn: CachingCdn, importMapResolver: ImportMapResolver) {
     this._cdn = cdn;
     this._importMapResolver = importMapResolver;
   }
 
+  _createLanguageService() {
+    // Initialize the language service and the host to manage it
+    // Files are added on the fly
+    this._languageServiceHost = new WorkerLanguageServiceHost(
```
aomarks, line 52: Maybe simpler to do this initialization on construction using a class field initializer? E.g.:

```ts
  private readonly _languageServiceHost = new WorkerLanguageServiceHost(
    self.origin,
    compilerOptions
  );

  private readonly _languageService = ts.createLanguageService(
    this._languageServiceHost,
    ts.createDocumentRegistry()
  );
```

src/typescript-worker/typescript.ts
```diff
@@ -145,19 +169,73 @@ export class TypeScriptBuilder {
...
+  /**
+   * Sync up the freshly acquired project files.
+   * In the syncing process files yet to be added are added, and versioned.
+   * Files that existed already but are modified are updated, and their version number
+   * gets bumped fo that the languageservice knows to update these files.
+   * */
+  sync(files: Map<string, string>) {
+    files.forEach((file, fileName) =>
+      this.updateFileContentIfNeeded(fileName, file)
+    );
```
aomarks, line 220: What happens if a file is deleted? Maybe we should iterate over the current `files`, and remove the ones that don't exist anymore?

I wonder if that will work properly in this edge case?:

1. compile with foo at version 0
2. delete foo and compile without it
3. re-create foo with different content, but it gets version 0 again because its new

aomarks (changes requested):
This is looking fantastic! This should improve compilation performance a lot right away, and sets things up nicely to be able to query it directly for completions etc.!

I have a few higher level comments, and then I'll do another round with some very minor style stuff.

But generally looking great! I think we should merge this as an isolated change once it's ready, since it already should have a big benefit.

src/playground-project.ts
```diff
@@ -545,6 +545,7 @@ export class PlaygroundProject extends LitElement {
       return;
     }
     workerApi.compileProject(
+      this._sessionId,
```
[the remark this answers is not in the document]
Matsuuu: Ah yeah that makes sense. I didn't have the full image of the project, and which layers are shared and which have their own instance.

If this is the case, I'll just remove the sessionid param and modify the caching into just a module scoped instance

src/typescript-worker/typescript.ts
Matsuuu: You're totally right. I did pass the files into the function earlier, but as I removed that and moved the file passing elsewhere, this can now be done in the constructor.

src/typescript-worker/typescript.ts
Matsuuu: Oh good point! I'll need to cover the deletion case. 

.. What comes to the deletion and re-addition, I'll need to check out if the LS is capable of handling it out of the box, or if we need to add some flags etc.

Matsuuu:
Cheers on the review! I implemented the changes proposed, and made the languageservice aware of removed files. I tested it out with re-naming, re-creating the file, etc. and didn't notice any hiccups.

If we merge this, I'll next start working on building the LSP functionalities

src/typescript-worker/typescript.ts
```diff
@@ -145,19 +154,82 @@ export class TypeScriptBuilder {
...
+    this.removeDeletedFiles(files);
+  }
+
+  removeDeletedFiles(files: Map<string, string>) {
```
aomarks, line 209:
```suggestion
  private _removeDeletedFiles(files: Map<string, string>) {
```

src/typescript-worker/typescript.ts
```diff
@@ -145,19 +154,82 @@ export class TypeScriptBuilder {
...
+    this.getScriptFileNames().forEach((fileName) => {
+      if (!files.has(fileName)) {
+        this.removeFile(fileName);
+      }
+    });
+  }
+
+  addFile(fileName: string, file: VersionedFile) {
```
aomarks, line 217:
```suggestion
  private _addFile(fileName: string, file: VersionedFile) {
```

Could also just consider moving its body directly into `updateFileContentIfNeeded`, because it's not used anywhere else (at least, after my other suggestion to call `updateFileContentIfNeeded` instead of `addFile` for the `d.ts` files).

src/typescript-worker/typescript.ts
```diff
@@ -145,19 +154,82 @@ export class TypeScriptBuilder {
...
+    this.files.set(fileName, file);
+  }
+
+  addFiles(files: Map<string, VersionedFile>) {
+    files.forEach((file, fileName) => {
+      this.files.set(fileName, file);
+    });
+  }
+
+  removeFile(fileName: string) {
```
aomarks, line 227:
```suggestion
  private _removeFile(fileName: string) {
```

Again, could also consider just moving the body directly into `updateFileContentIfNeeded` since it's only used in one place, and is very simple self-describing code already.

src/typescript-worker/typescript.ts
```diff
@@ -145,19 +154,82 @@ export class TypeScriptBuilder {
...
+    this.files.delete(fileName);
+  }
+
+  _fileContentHasUpdated(oldFile: VersionedFile, newContent: string) {
```
aomarks, line 231:
```suggestion
  private _fileContentHasUpdated(oldFile: VersionedFile, newContent: string) {
```

Here too -- it's pretty simple self-documenting code only used in one place, so you could just do the check directly in `updateFileContentIfNeeded`.

src/typescript-worker/build.ts
```diff
@@ -14,6 +13,7 @@ import type {
   BuildOutput,
 } from '../shared/worker-api.js';
 import {CachingCdn} from './caching-cdn.js';
+import { getBuilder } from './project-cache.js';
```
aomarks, line 16: Very minor, but mind running `npm run format` to Prettier-format everything?

aomarks (changes requested):
Nice! Looking good. Two comments that affect the behavior, everything else just some style suggestions.

Would you also mind adding an entry to the `CHANGELOG.md` file? Something like

"TypeScript files are now compiled incrementally for improved performance."

aomarks:
> Would you also mind adding an entry to the `CHANGELOG.md` file? Something like

Oh also I just recently merged https://github.com/PolymerLabs/playground-elements/pull/204 which upgrades to Lit 2 and updates the CHANGELOG, so you might want to `merge main` before you modify the CHANGELOG, because otherwise we'll probably have to do a merge conflict resolve afterwards.

src/typescript-worker/project-cache.ts
```diff
@@ -3,13 +3,12 @@ import {CachingCdn} from './caching-cdn.js';
 import { ImportMapResolver } from "./import-map-resolver";
 
 
-export const builderCache = new Map();
+export let builderInstance: TypeScriptBuilder;
```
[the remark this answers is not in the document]
Matsuuu: If we are going to implement more LSP support (completion, types, etc.), we need a LS instance for those too.

I was wondering if we should/could use the same `TypeScriptBuilder` instance for those too, since it will be up to date with the latest project files? If we do that, we might want to keep the `project-cache.ts` separate, and maybe even re-name the builder to something else.

Thoughts?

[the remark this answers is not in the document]
aomarks: Good point, it will probably make sense to have some separation for the reason you said. And yeah, I also agree it will probably make sense to rename it -- or maybe we'll want to factor out another class that purely encapsulate the language service parts, which is passed *in* to the builder. Something to consider in the near future!

Only thing we should do is add a license header to this file, then:

```ts
/**
 * @license
 * Copyright 2021 Google LLC
 */
```

Matsuuu:
Cheers on the thorough review once again @aomarks !

I've done the changes, applied formatting and coding styles accordingly, and removed reduntant functions.

I also merged the upstream main and updated the changelog. 

After we merge this, I'll start working on a plan on re-structuring the typescriptworker for the LSP functionality (in a different branch / pr of course).

aomarks (approved):

aomarks:
A few simple lint errors in https://github.com/PolymerLabs/playground-elements/pull/205/checks?check_run_id=3653621020 to fix, then ready to merge.

Matsuuu:
Ahh duh. Of course I forgot to lint :D 

Fixed it now!

aomarks:
> Ahh duh. Of course I forgot to lint :D
> 
> Fixed it now!

No worries.

I think the formatting has reverted again, indentation looks different, could you also just run `npm run format`?

Matsuuu:
Sorry for the inconvenience. As I said earlier, still kinda new to these larger projects and contributing :) Should be fine now

aomarks:
> Sorry for the inconvenience. As I said earlier, still kinda new to these larger projects and contributing :) Should be fine now

No problem!

Issue #203, opened 2021-09-17
Codecompletion


Matsuuu:
Heya,

Awesome work so far with the playground elements. I'd love to contribute to this as I'll most likely be adopting playground-elements as the code editing/inspection platform for [Web Component DevTools](https://github.com/Matsuuu/web-component-devtools).

I was wondering if there is plans, or if there is a reason there's no code completion in playground elements? I know it can be a bit heavier to implement, but seeing that [Codemirror has examples on how to do it](https://codemirror.net/demo/complete.html), it would be possible.

If not built in, would there be interest in a addon to enable intellisense/autocompletion to playgrounds?

I'd be happy to work on that since it's something I'd want in some of my own personal projects too

aomarks:
> Heya,
> 
> Awesome work so far with the playground elements. I'd love to contribute to this as I'll most likely be adopting playground-elements as the code editing/inspection platform for [Web Component DevTools](https://github.com/Matsuuu/web-component-devtools).

Awesome!

> I was wondering if there is plans, or if there is a reason there's no code completion in playground elements? I know it can be a bit heavier to implement, but seeing that [Codemirror has examples on how to do it](https://codemirror.net/demo/complete.html), it would be possible.

No particular reason it hasn't been worked on yet, just lack of time among many other priorities.

> If not built in, would there be interest in a addon to enable intellisense/autocompletion to playgrounds?
> 
> I'd be happy to work on that since it's something I'd want in some of my own personal projects too

Yeah, I think that would be great!

(Personally the thing I miss even more than code-completion coming from VSCode is getting documentation/types when I hover over something. I think that's all part of the same family of intellisense-related queries to the TypeScript language server, though, so if we can get one working, the other should be simple too.)

I think the main obstacle to getting this working is going to be on the worker side. We currently create a new TypeScript language service instance for every compile (https://github.com/PolymerLabs/playground-elements/blob/e5cb8462fbdfa1238b919705d4738a795b5453a9/src/typescript-worker/typescript.ts#L88). What we should do instead is have a persistent language service per project worker, and keep track of which files have changed between compiles. I believe we could then query this persistent language service (we'd need a new <playground-project> -> worker API for this) for intellisense-related info. This would also enable incremental compilation, which would be very beneficial for compilation performance.

PR #211, opened 2021-09-21
Refactor the LanguageService into a separate module

Matsuuu:
- Removed builder parts from typescript.ts and moved them to build.ts
- Made renamed TypeScriptBuilder to LanguageServiceContext
- Changed Language Service functions to accept LS context as parameter
- Made use of project-cache easier by providing sane defaults

This refactoring is done in part as preparation for features like [this](https://github.com/google/playground-elements/issues/203)


src/typescript-worker/build.ts
```diff
@@ -5,30 +5,134 @@
...
+      files.map((file) => ({kind: 'file', file}))
+    )
   );
   for await (const result of results) {
     emit(result);
   }
   emit({kind: 'done'});
 };
+
+async function* processFiles(
```
aomarks, line 43: I think this function should be in a file called `typescript-builder.ts`

src/typescript-worker/language-service-context.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Compiles a project, returning a Map of compiled file contents. The map only
+ * contains context for files that are compiled. Other files are skipped.
+ *
+ * TODO (justinfagnani):  We could share a DocumentRegistry across
+ * multiple <playground-project> instances to save memory and type analysis of
+ * common lib files like lit-element, lib.d.ts and dom.d.ts.
+ */
+export class LanguageServiceContext {
+  public readonly compilerOptions = compilerOptions;
+  public readonly cdn: CachingCdn;
```
aomarks, line 32: It seems like `CachingCdn` and `ImportMapResolver` probably don't quite belong in this class, because they're not very closely tied to the TypeScript/LanguageService: [1] they're used by the `BareModuleBuilder` which isn't TypeScript/LanguageService-related and [2] none of the methods here actually need access to the `CachingCdn` or `ImportMapResolver`

How about factoring out one more class -- something like this?:

- Rename `project-cache.ts` to `worker-context.ts`
- Add a new class called `WorkerContext` inside `worker-context.ts`
- The new class has these fields:
  - `CachingCdn`
  - `ImportMapResolver`
  - `LanguageServiceContext`
- Rename `getLanguageServiceContext` to `getWorkerContext`.

src/typescript-worker/language-service-context.ts
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Compiles a project, returning a Map of compiled file contents. The map only
+ * contains context for files that are compiled. Other files are skipped.
+ *
+ * TODO (justinfagnani):  We could share a DocumentRegistry across
+ * multiple <playground-project> instances to save memory and type analysis of
+ * common lib files like lit-element, lib.d.ts and dom.d.ts.
+ */
+export class LanguageServiceContext {
+  public readonly compilerOptions = compilerOptions;
```
aomarks, line 31: Nit: omit the `public` keyword, since it's the default.

src/typescript-worker/project-cache.ts
```diff
@@ -4,30 +4,41 @@
...
+ *
+ * You can provide the caching cdn and moduleResolved instances yourself
+ * or let the function generate then from default values.
  */
-export function getBuilder(
+export function getLanguageServiceContext(
   config: TypeScriptWorkerConfig,
-  cdn: CachingCdn,
-  moduleResolver: ImportMapResolver
+  cdn?: CachingCdn,
```
aomarks, line 25: Since we're in a refactoring mood...

I think this function should *only* take a `config`. If there's a cache key hit, we already have a `CachingCdn` and `ImportMapResolver` in the stored context. Otherwise we create new ones from the config.

src/typescript-worker/project-cache.ts
```diff
@@ -4,30 +4,41 @@
...
  * Acquire the existing builder instance, or create a fresh one if missing.
  * If the config differs from the existing instance's config, a new TypeScriptBuilder is
  * instantiated and made the new builderInstance.
+ *
+ * You can provide the caching cdn and moduleResolved instances yourself
+ * or let the function generate then from default values.
  */
-export function getBuilder(
+export function getLanguageServiceContext(
   config: TypeScriptWorkerConfig,
```
aomarks, line 24: How about we rename `TypeScriptWorkerConfig` to simply `WorkerConfig` -- because not everything is TypeScript related (the CDN and ImportMap is also used for the bare module transformer, which doesn't use TypeScript).

(It's true that all of this code is under `typescript-worker/` for historical reasons, but the `BareModuleBuilder` doesn't use TypeScript, and `.js` files flow through it without hitting any TypeScript code. So I've started thinking about the code in this directory a little more broadly -- it's concerned with everything that happens in the web worker, including but not limited to TypeScript. Probably we should eventually rename this directory to just `web-worker/` -- but that's for some other time.)

aomarks (changes requested):
Great start!

Matsuuu:
Thanks for the thorough review again! Great points and ideas all around. I'll try to get around implementing those during this week.

Excited to continue work on this

Matsuuu:
I'm doing the refactor of the `TypeScriptWorker` => `Worker` at the moment and will be listing a few things we need to keep in mind when we move the full naming away from TS.

- Rename the `typescript-worker` -directory to just `web-worker`
- Change the build output in RollUp to match (`playground-typescript-worker.js` => `playground-web-worker.js` )
- Change the worker file references in `playground-project.ts`
- Change the variable naming to remove typescript from it.


The current refactor in this branch has already taken the steps to rename the `TypeScriptWorkerAPI` to `WorkerAPI` and `TypeScriptWorkerConfig` to `WorkerConfig`.

Matsuuu:
Implemented the review changes and did some re-naming. Listed in the above comment are some points we can maybe take into action in this PR or after this PR in a separate one, as it's a bigger refactor, containing packaging and rollup stuff

src/typescript-worker/typescript-builder.ts
```diff
@@ -0,0 +1,111 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {BuildOutput, SampleFile} from '../shared/worker-api.js';
+import {TypesFetcher} from './types-fetcher.js';
+import {PackageJson} from './util.js';
+import {makeLspDiagnostic} from './diagnostic.js';
+import {WorkerContext} from './worker-context.js';
+
+export async function* processFiles(
```
aomarks, line 13: How about `processTypeScriptFiles`

src/typescript-worker/language-service-context.ts
```diff
@@ -0,0 +1,130 @@
...
+    ts.createDocumentRegistry()
+  );
+}
+
+export interface VersionedFile {
+  version: number;
+  content: string;
+}
+
+export class WorkerLanguageServiceHost implements ts.LanguageServiceHost {
```
aomarks, line 46:
```suggestion
class WorkerLanguageServiceHost implements ts.LanguageServiceHost {
```

src/typescript-worker/language-service-context.ts
```diff
@@ -0,0 +1,130 @@
...
+    compilerOptions
+  );
+
+  readonly service = ts.createLanguageService(
+    this.serviceHost,
+    ts.createDocumentRegistry()
+  );
+}
+
+export interface VersionedFile {
```
aomarks, line 41:
```suggestion
interface VersionedFile {
```

src/typescript-worker/worker-context.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LanguageServiceContext} from './language-service-context.js';
+import {CachingCdn} from './caching-cdn.js';
+import {ImportMapResolver} from './import-map-resolver.js';
+import {WorkerConfig} from '../shared/worker-api.js';
+
+export let workerContext: WorkerContext;
```
aomarks, line 12:
```suggestion
let workerContext: WorkerContext | undefined;
```

src/typescript-worker/worker-context.ts
```diff
@@ -0,0 +1,44 @@
...
+let cacheKey = '';
+
+/**
+ * Acquire the existing worker instance, or create a fresh one if missing.
+ * If the config differs from the existing instance's config, a new WorkerContext is
+ * instantiated and made the new workerInstance.
```
aomarks, line 18:
```suggestion
 * instantiated and made the new instance.
```

src/typescript-worker/worker-context.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * You can provide the caching cdn and moduleResolved instances yourself
```
aomarks, line 20:
```suggestion
```

src/typescript-worker/worker-context.ts
```diff
@@ -0,0 +1,44 @@
...
+ * or let the function generate then from default values.
```
aomarks, line 21:
```suggestion
```

aomarks (changes requested):
This is great! Really cleans up the code.

Just a few tiny nits in some names/comments, otherwise LGTM.

aomarks:
> Implemented the review changes and did some re-naming. Listed in the above comment are some points we can maybe take into action in this PR or after this PR in a separate one, as it's a bigger refactor, containing packaging and rollup stuff

Sounds good! Yeah, let's do the other renames in a separate PR.

> * Change the build output in RollUp to match (`playground-typescript-worker.js` => `playground-web-worker.js` )

Note this will be a breaking change, because some users will have hard-coded this name into their build process (e.g. we do so for lit.dev). I'm ok with that. but pretty soon I think we should release 1.0.0 and then be a bit more conservative about breaking changes.

Matsuuu:
Allright. This should be allright to merge now.

I agree about being extra careful on doing anything breaking. We can create a separate issue about it and plan on how to ship the changes

aomarks (approved):

PR #214, opened 2021-09-24
Fix: Add property bindings to playground-code-editor's value

### LINKED ISSUE #213: [Bug]: Playground code editor doesn't allow values as initial params
Codesandbox: https://codesandbox.io/s/funny-kalam-kts6h?file=/src/index.js

The [Playground code editor](https://github.com/google/playground-elements#playground-code-editor) API documentation says that you can give the code editor a value as a property. This however is not possible through means like like, like seen in the codesandbox above.

I'm quite sure it's due to the way we handle values in the code editor. [The value property is not within the Lit's reactive properties](https://github.com/google/playground-elements/blob/main/src/playground-code-editor.ts#L126-L138) but as custom setter.

We should have some kind of way to set the value for a code editor, when you are not using any other parts of the playgrounds elements.

Matsuuu:
Fixes #213 

Do you see a reason why we wouldn't want to have a property binding on the value field? We have it in our `updated` -method's switch cases so it would make sense to have it as a lit property.

There's the comment on the private value field, but the property would bind to the public getter/setter.

From my testing I didn't see stuff break, and the tests run fine, so I think this is just a missing binding and will fix the case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Upgrade to Lit 2.0.
 
+### Fixed
+
+- Fixed bug where the `value` property of a `<playground-code-editor>` did not take effect if set before element upgrade.
+
+### Added
+
+- It is now possible to set the `value` of a `<playground-code-editor>` using the `value` HTML attribute.
+
 ## [0.13.0] - 2021-09-14
 
 ### Changed
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -127,6 +127,7 @@ export class PlaygroundCodeEditor extends LitElement {
   // that we can set this value internally without triggering an update.
   private _value?: string;
 
+  @property()
   get value() {
     return this._value;
   }


aomarks (approved):
Cool, I think this makes sense. Adding `@property` here has 2 practical effects:

1. It will be possible to set the value by updating the `value` HTML attribute, whereas previously you could only use the property. Seems like it might be unusual to do this, but I don't see why not I guess? It would let you put some code in an editor from declarative HTML.

2. If `value` is set before the element upgrades, it will now pass through the setter and be handled correctly. This is the main benefit of this change, I think.

The 3rd effect of calling `update` when the value changes doesn't apply here, because we already define our own getter/setter that does this (this allows us to directly set the underlying `_value` property without triggering an update for when the code is changed from within by the user typing).

aomarks (changes requested):
Oh, would you mind updating the `CHANGELOG` file, with an entry like this:

- It is now possible to directly set the `value` of a `<playground-code-editor>` from an HTML attribute, and also before the element upgrades.

Matsuuu:
Updated the changelog.

`Seems like it might be unusual to do this, but I don't see why not I guess? `

I actually ran into this while working on a "console" window for Web Component Devtools.

I'm planning on creating a console similiar to the console in chrome devtools, and used the `playground-code-editor` as it's base since that's really the only part I need for it.

aomarks:
> I'm planning on creating a console similiar to the console in chrome devtools, and used the `playground-code-editor` as it's base since that's really the only part I need for it.

Ah, cool! This makes me glad that this library is factored out into small components :)

aomarks (approved):

PR #218, opened 2021-10-01
More aggressive techniques for updating the service worker

aomarks:
More aggressive techniques to encourage new versions of the service worker to be found and activated:

  - The expected version number is now included as a URL parameter to encourage cache busting (e.g. `playground-service-worker.js?v=0.14.3`).

  - The previous version of a service worker is now explicitly unregistered, to work around cases where a previous version would never shut down to allow a new version to take over, even across browser restarts.

  - The previous and expected new service worker version numbers are now included in console log messages, to aid in debugging.

### DIFF
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -219,9 +219,7 @@ export class PlaygroundProject extends LitElement {
   @state()
   private _serviceWorkerAPI?: Remote<ServiceWorkerAPI>;
 
-  private _deferredTypeScriptWorkerApi = new Deferred<
-    Remote<WorkerAPI>
-  >();
+  private _deferredTypeScriptWorkerApi = new Deferred<Remote<WorkerAPI>>();
 
   private _validImportMap: ModuleImportMap = {};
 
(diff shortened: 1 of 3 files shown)


src/playground-project.ts
```diff
@@ -479,7 +481,9 @@ export class PlaygroundProject extends LitElement {
           // immediately. We'll get back here again after it updates via a
           // CONNECT_PROJECT_TO_SW message from the proxy.
           console.info(
-            'Playground service worker is outdated, waiting for update.'
+            `Playground service worker is outdated. ` +
+              `Want ${version} but got ${e.data.version}. ` +
+              `Waiting for update.`
```
justinfagnani, line 486: is there a corresponding message after this that the update happened?

justinfagnani (approved):

src/playground-project.ts
aomarks: To show in the console? No, but the playground will start working so it's pretty obvious.

PR #219, opened 2021-10-02
Fix bottleneck in build causing significant preview latency

aomarks:
Fixes an oversight in the `MergedAsyncIterables` class that caused most build output to get blocked on TypeScript semantic (type checker) errors being computed. This has a very noticeable effect on responsiveness.

Check out the difference (e.g. modify the "Hello..." template):

Before: https://lit.dev/playground/
After: https://pr543-7afcc35---lit-dev-5ftespv5na-uc.a.run.app/playground/

Also threw in a `NO_TERSER` mode for Rollup, for slightly faster dev.

### DIFF
--- a/src/typescript-worker/util.ts
+++ b/src/typescript-worker/util.ts
@@ -5,14 +5,15 @@
  */
 
 /**
- * Merges multiple async iterables into one iterable. Iterator next promises are
- * raced, so order is not preserved. Additional iterables can be added before or
- * during iteration.
+ * Merges multiple async iterables into one iterable. Order is not preserved.
+ * Iterables can be added before or during iteration. After exhausted, adding a
+ * new iterator throws.
  */
 export class MergedAsyncIterables<T> {
   private readonly _buffer: Array<{value: T; emitted: () => void}> = [];
   private _numSources = 0;
   private _notify?: () => void;
+  private _done = false;
 
   async *[Symbol.asyncIterator]() {
     while (this._numSources > 0) {
@@ -26,19 +27,27 @@ export class MergedAsyncIterables<T> {
       await new Promise<void>((resolve) => (this._notify = resolve));
       this._notify = undefined;
     }
+    this._done = true;
   }
 
-  async add(iterable: AsyncIterable<T>) {
-    this._numSources++;
-    for await (const value of iterable) {
-      // Wait for this value to be emitted before continuing
-      await new Promise<void>((emitted) => {
-        this._buffer.push({value, emitted});
-        this._notify?.();
-      });
+  add(iterable: AsyncIterable<T>) {
+    if (this._done) {
+      throw new Error(
+        'Merged iterator is exhausted. Cannot add new source iterators.'
+      );
     }
-    this._numSources--;
-    this._notify?.();
+    this._numSources++;
+    (async () => {
+      for await (const value of iterable) {
+        // Wait for this value to be emitted before continuing
+        await new Promise<void>((emitted) => {
+          this._buffer.push({value, emitted});
+          this._notify?.();
+        });
+      }
+      this._numSources--;
+      this._notify?.();
+    })();
   }
 }
 
--- a/src/typescript-worker/util.ts
+++ b/src/typescript-worker/util.ts
@@ -4,58 +4,41 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {Deferred} from '../shared/deferred.js';
-
-type IteratorResultWithNext<T> = Promise<{
-  value: IteratorResult<T>;
-  next: () => IteratorResultWithNext<T>;
-}>;
-
-const NEW_ITERATOR = Symbol();
-
 /**
  * Merges multiple async iterables into one iterable. Iterator next promises are
  * raced, so order is not preserved. Additional iterables can be added before or
  * during iteration.
  */
 export class MergedAsyncIterables<T> {
-  private _iterators = new Map<
-    () => IteratorResultWithNext<T>,
-    IteratorResultWithNext<T>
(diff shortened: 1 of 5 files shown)


src/typescript-worker/util.ts
```diff
@@ -27,11 +32,23 @@ export class MergedAsyncIterables<T> {
...
   }
 
   async *[Symbol.asyncIterator]() {
     while (this._iterators.size > 0) {
-      const {value, next} = await Promise.race(this._iterators.values());
+      // Emit any available value from our source iterators, or notice there is
+      // a new iterator. We should immediately add new iterators to this race,
+      // because a new one might have a value before the older ones.
+      const result = await Promise.race([
+        ...this._iterators.values(),
```
justinfagnani, line 44: This definitely works and should go in, but I feel like it'd be a bit more straight-forward with an approach that iterates on each iterable directly. Essentially combining an async iteratable that let's you push new individual values in (like TestAsyncIterable in lit-html tests) with this...

```ts
add(iterable: AsyncIterable<T>) {
  for await (const i of iterable) {
    await this.push(i);
  }
}
```
...to push in new values. I haven't tested, but here's the idea:

```ts
export class MergedAsyncIterables<T> {
  private _buffer: Array<{value: T, emitted: () => void}> = [];  
  private _notifyNewValue?: () => void;
    
  async *[Symbol.asyncIterator]() {
    while (true) {
      if (this._buffer.length > 0) {
        const {value, emitted} = this._buffer.shift();
        yield value;
        // Let the loop in add() continue
        emitted();
      } else {
        // Wait for a new value
        await new Promise<void>((resolve) => (this._notifyNewValue = resolve));
        this._notifyNewValue = undefined;
      }
    }
  }
  
  async add(iterable: AsyncIterable<T>) {
    for await (const i of iterable) {
      // Wait for this value to be emitted before continuing
      await this._push(i);
    }
  }
  
  private async _push(value: T): Promise<void> {
    let emitted: () => void;
    const emittedPromise = new Promise<void>((res) => emitted = res);
    this._buffer.push({value, emitted});
    this._notifyNewValue?.();
    // Let the loop in add() wait for this value to be emitted
    return emittedPromise;
  }
}
```

justinfagnani (approved):

src/typescript-worker/util.ts
```diff
@@ -27,11 +32,23 @@ export class MergedAsyncIterables<T> {
       next,
     });
     this._iterators.set(next, next());
+    this._newIterator.resolve(NEW_ITERATOR);
```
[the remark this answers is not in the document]
aomarks: Yes we add them dynamically. This class is used in https://github.com/google/playground-elements/blob/main/src/typescript-worker/bare-module-transformer.ts. We create this merged async iterable, and we add a new iterator for each module as we walk through the dependency graph.

src/typescript-worker/util.ts
aomarks: Nice! That is simpler.

The only change I had to make was keeping track of how many active source iterators there were, and also notifying when one is exhausted, otherwise we never know when we are done. Slightly modified some other code too, but only the one change to the logic.

PTAL

src/typescript-worker/util.ts
```diff
@@ -4,41 +4,41 @@
...
 
   async *[Symbol.asyncIterator]() {
-    while (this._iterators.size > 0) {
-      const {value, next} = await Promise.race(this._iterators.values());
-      if (value.done) {
-        this._iterators.delete(next);
-      } else {
-        yield value.value;
-        this._iterators.set(next, next());
+    while (this._numSources > 0) {
```
justinfagnani, line 18: There is a race condition here where if existing iterators are done, adding an iterator will not do anything. I'd try to add an error in that case below, and either make this class take a required initial iterator in the constructor (just call `this.add()`) or wait for at least one value by changing this to a do/while loop so the `await` kicks in.

src/typescript-worker/util.ts
```diff
@@ -4,41 +4,41 @@
...
+        emitted();
       }
+      // Wait until there is a new value or a source iterator was exhausted
+      await new Promise<void>((resolve) => (this._notify = resolve));
+      this._notify = undefined;
+    }
+  }
+
+  async add(iterable: AsyncIterable<T>) {
+    this._numSources++;
```
justinfagnani, line 32: If you change to a do/while above, a good check here would be:

```ts
if (this._numSources === 0 && this._notify === undefined) {
  throw ...
}
```

src/typescript-worker/util.ts
```diff
@@ -4,41 +4,41 @@
...
+    for await (const value of iterable) {
+      // Wait for this value to be emitted before continuing
+      await new Promise<void>((emitted) => {
```
justinfagnani, line 35: Nice inlining of push()!

src/typescript-worker/util.ts
aomarks: I like the error if you try to add after iteration is complete, done!

Not sure the constructor idea could work, because then you can't do this (which is what I'm doing):

```ts
    const output = new MergedAsyncIterables<BuildOutput>();
    output.add(this._handleProjectFiles(results, output));
```

Also not sure about making it always wait for an initial value, because it feels illogical to me that this pattern would never resolve. This is actually one of the existing unit tests. It seems more logical to me that if you iterate over a merged set of no iterables, you successfully get no values, rather than a deadlock.

```ts
    const merged = new MergedAsyncIterables();
    for await (const v of merged) {
    }
```

justinfagnani: Sounds good!

justinfagnani (approved):

Issue #220, opened 2021-10-02
Reloading preview adds history states


aomarks:
When an iframe reloads, the parent window's history is affected. So new history states are created whenever a playground preiew reloads. I think the trick to prevent that would be to remove the iframe from the DOM before we change its src, and then attach it after.

https://github.com/google/playground-elements/blob/05538da0ee6fe13b3345d89b805dd06b2b8f30b3/src/playground-preview.ts#L198

PR #221, opened 2021-10-02
Fix reloading preview iframe affecting parent window history

aomarks:
Fixes a bug where every time the preview reloaded, a history entry would be added to the parent window. From my testing, this affects Chrome but not Firefox, and only when the parent/iframe origins are separate. Removing the iframe from the DOM before doing the reload, and then adding it back again, prevents this from happening.

Fixes https://github.com/google/playground-elements/issues/220

cc @castastrophe @Matsuuu

### DIFF
--- a/src/playground-preview.ts
+++ b/src/playground-preview.ts
@@ -256,10 +256,8 @@ export class PlaygroundPreview extends PlaygroundConnectedElement {
     // window (on Chrome but not Firefox, and only when the parent/iframe origins
     // are different). Removing the iframe from the DOM while we initiate the
     // reload prevents a history entry from being added.
-    const parentNode = iframe.parentNode;
-    const nextSibling = iframe.nextSibling;
+    const {parentNode, nextSibling} = iframe;
     if (parentNode) {
-      iframe.nextSibling;
       iframe.remove();
     }
     // Note we can't use contentWindow.location.reload() here, because the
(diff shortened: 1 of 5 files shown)


src/playground-preview.ts
```diff
@@ -248,13 +248,27 @@ export class PlaygroundPreview extends PlaygroundConnectedElement {
...
       return;
     }
+    // Reloading the iframe can cause a history entry to be added to the parent
+    // window (on Chrome but not Firefox, and only when the parent/iframe origins
+    // are different). Removing the iframe from the DOM while we initiate the
+    // reload prevents a history entry from being added.
+    const parentNode = iframe.parentNode;
+    const nextSibling = iframe.nextSibling;
+    if (parentNode) {
+      iframe.nextSibling;
```
justinfagnani, line 262: what's this?

aomarks: Oops, a leftover. Removed.

src/playground-preview.ts
```diff
@@ -248,13 +248,27 @@ export class PlaygroundPreview extends PlaygroundConnectedElement {
...
+    const iframe = this._iframe;
+    if (!iframe) {
       return;
     }
+    // Reloading the iframe can cause a history entry to be added to the parent
+    // window (on Chrome but not Firefox, and only when the parent/iframe origins
+    // are different). Removing the iframe from the DOM while we initiate the
+    // reload prevents a history entry from being added.
+    const parentNode = iframe.parentNode;
+    const nextSibling = iframe.nextSibling;
```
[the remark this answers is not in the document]
aomarks: Done.

justinfagnani (approved):

PR #222, opened 2021-10-04
Use service worker hash instead of NPM version to check staleness

aomarks:
Previously we used the NPM version to determine when we needed to force a service worker update. Now we use a hash of the service worker itself. This should cause much less churn going forward, since most releases don't modify the service worker bytes.

Also rolls back one of the changes in https://github.com/google/playground-elements/pull/218 -- adding a `?v=` query param to the service worker URL -- because unpkg.com (the default URL for the service worker) serves a redirect to strip off any query parameters you use, and service worker script responses are not allowed to redirect (see https://w3c.github.io/ServiceWorker/#:~:text=Set%20request%E2%80%99s%20redirect%20mode%20to%20%22error%22).


justinfagnani:
> service worker script responses are not allowed to redirect

TIL

scripts/update-version-module.js
```diff
@@ -5,23 +5,60 @@
...
+};
 
-  const newVersionModuleCode = `/**
+const getServiceWorkerHash = async () => {
+  const path = join(thisDir, '..', 'playground-service-worker.js');
+  const code = await fs.readFile(path, 'utf8');
+  // Note in theory Terser could minimize the version code differently, but it
+  // should be obvious because this script will start to fail.
+  const versionPattern =
+    /version:\s*(?<quote>["'`])(?<hex>[a-z0-9]+)\k<quote>/g;
```
justinfagnani, line 35: `\k<quote>` whoa. I didn't know named capture groups back-references yet!

scripts/update-version-module.js
```diff
@@ -31,15 +68,19 @@ import {fileURLToPath} from 'url';
...
-  const oldVersionModuleCode = await fs.readFile(versionModulePath, 'utf8');
-  if (oldVersionModuleCode !== newVersionModuleCode) {
-    await fs.writeFile(versionModulePath, newVersionModuleCode, 'utf8');
-    console.log('Updated src/shared/version.ts, please commit this change.');
-    // Fail to help the publisher remember to commit.
-    process.exit(1);
-  }
-})();
+const versionModulePath = join(thisDir, '..', 'src', 'shared', 'version.ts');
+const oldVersionModuleCode = await fs.readFile(versionModulePath, 'utf8');
```
justinfagnani, line 76: fwiw, there's three i/o operations here that could be parallelized here if there were lifted up to a `Promise.all()`

justinfagnani (approved):

scripts/update-version-module.js
aomarks: Yeah, but this runs so rarely and is so fast that I'd rather just keep the code as simple as possible.

PR #223, opened 2021-10-04
Housekeeping and CI improvements

aomarks:
- Adds `format`, `lint`, and `version-module` CI jobs
- Upgrade CI node version. Add caching. Use new recommendation for installing playwright deps.
- Upgrade dependencies
- Fix format and lint errors
- Misc cleanup


.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
+name: lint
```
justinfagnani, line 1: It it that useful to have format and lint as separate workflows? Is it just so both run if one fails?

justinfagnani (approved):

.github/workflows/lint.yml
aomarks: Yeah it's so that the checkmarks are really easy to read. One for each thing that could fail. A little wasteful but :shrug:

Issue #78, opened 2020-11-11
Warn when service worker is not available


aomarks:
We should do this visibly in the element, not in the console.

Note that in Firefox, Service Workers are always disabled in private browsing mode.

aomarks:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1320796 for Firefox issue about service workers not being available in incognito mode.

PR #224, opened 2021-10-04
Show useful error when service workers unavailable

aomarks:
Detect when service workers are not available and show a useful error. Uses the same styling that we previously used on the main editor to prevent keyboard focus trap. The main place this comes up is when using Firefox in private browsing mode, which is not currently supported (see https://bugzilla.mozilla.org/show_bug.cgi?id=1320796).

In a followup I'm also going to use this style of overlay to inform the user that something has gone wrong with service worker update, and that they may need to reload the page.

Fixes https://github.com/google/playground-elements/issues/78

### DIFF
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -9,6 +9,7 @@ import {customElement, property, query, state} from 'lit/decorators.js';
 import {ifDefined} from 'lit/directives/if-defined.js';
 import {CodeMirror} from './internal/codemirror.js';
 import playgroundStyles from './playground-styles.js';
+import './internal/overlay.js';
 import type {Diagnostic} from 'vscode-languageserver';
 
 // TODO(aomarks) Could we upstream this to lit-element? It adds much stricter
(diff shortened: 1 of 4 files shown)


src/playground-code-editor.ts
```diff
@@ -255,12 +236,12 @@ export class PlaygroundCodeEditor extends LitElement {
         @keydown=${this._onKeyDown}
       >
         ${this._showKeyboardHelp
-          ? html`<div id="keyboardHelpScrim">
+          ? html`<playground-internal-overlay>
```
justinfagnani, line 239: btw, I don't think this text is very descriptive about what's going on or why the popup is showing. For non screen-reader users there appears to be no action to take. Maybe add some descriptive text about how you can now use <tab> to move focus to other parts of the page?

Also, once you've hit escape to trigger the popup, do you have to hit it against the "exit" the editor? Or is it already in that state? Hitting escape again doesn't appear to do anything (in the playground).

src/playground-preview.ts
```diff
@@ -166,7 +210,7 @@ export class PlaygroundPreview extends PlaygroundConnectedElement {
         </mwc-icon-button>
       </div>
 
-      <div id="content">
+      <div id="content" class=${classMap({error: !!this._error})}>
```
justinfagnani, line 213: `classMap` already works on truthiness... is this to quiet the type-checker?

justinfagnani (approved):

src/playground-preview.ts
aomarks: Yeah, `Type 'undefined' is not assignable to type 'string | number | boolean'`

src/playground-code-editor.ts
aomarks: Filed https://github.com/google/playground-elements/issues/225 with some thoughts, PTAL.

PR #227, opened 2021-10-04
Expose cursor position and token under cursor from editor

Matsuuu:
Exposed useful information for diagnostic and language server interactions from the code-editor to the outside.

### DIFF
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -11,6 +11,7 @@ import {CodeMirror} from './internal/codemirror.js';
 import playgroundStyles from './playground-styles.js';
 import './internal/overlay.js';
 import type {Diagnostic} from 'vscode-languageserver';
+import type {Position} from 'codemirror';
 
 // TODO(aomarks) Could we upstream this to lit-element? It adds much stricter
 // types to the ChangedProperties type.
@@ -21,6 +22,15 @@ interface TypedMap<T> extends Map<keyof T, unknown> {
   keys(): IterableIterator<keyof T>;
   values(): IterableIterator<T[keyof T]>;
   entries(): IterableIterator<{[K in keyof T]: [K, T[K]]}[keyof T]>;
+}
+
+export interface EditorToken {
+  /** The character (on the given line) at which the token starts. */
+  start: number;
+  /** The character at which the token ends. */
+  end: number;
+  /** Code string under the cursor. */
+  string: string;
 }
 
 const unreachable = (n: never) => n;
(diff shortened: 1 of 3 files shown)


src/playground-code-editor.ts
```diff
@@ -101,9 +111,36 @@ export class PlaygroundCodeEditor extends LitElement {
...
 
-  // Used by tests.
   protected _codemirror?: ReturnType<typeof CodeMirror>;
 
+  get cursorPosition(): Position | undefined {
+    const cm = this._codemirror;
+    if (!cm) return undefined;
+
+    return cm.getCursor('start');
+  }
```
aomarks, line 121:
```suggestion
  get cursorPosition(): Position {
    return this._codemirror?.getCursor('start') ?? {ch: 0, line: 0};
  }
```

src/playground-code-editor.ts
```diff
@@ -101,9 +111,36 @@ export class PlaygroundCodeEditor extends LitElement {
...
+
+  get cursorIndex(): number | undefined {
```
aomarks, line 123: Similar to above, maybe we should default to 0 if codemirror is uninitialized? Since it will always start there anyway?

src/playground-code-editor.ts
aomarks: There's also a `sticky` property in this codemirror `Position` type. It's not documented, do you know what it does? If not needed right now, maybe we should strip that (can always add it later, but seems like a codemirror-specific thing).

src/test/playground-ide_test.ts
```diff
@@ -494,6 +494,72 @@ suite('playground-ide', () => {
     assert.isFalse(ide.modified);
     assert.isFalse(ide.modified);
   });
+  test('returns the correct cursor position and index', async () => {
```
aomarks, line 497:
```suggestion

  test('returns the correct cursor position and index', async () => {
```

src/test/playground-ide_test.ts
```diff
@@ -494,6 +494,72 @@ suite('playground-ide', () => {
...
+      type: codeToAdd,
+    });
+
+    assert.equal(editor.value, codeToAdd);
+    assert.equal(editor.cursorIndex, codeToAdd.length);
+    const cursorPosition = editor.cursorPosition;
+    assert.equal(cursorPosition?.line, 1);
+    assert.equal(cursorPosition?.ch, 23);
+  });
+  test('returns the token under cursor', async () => {
```
aomarks, line 531:
```suggestion

  test('returns the token under cursor', async () => {
```

aomarks (changes requested):
Looks great!

A couple comments about defaults, nothing major.

Also mind also adding a CHANGELOG entry?

aomarks (approved):

PR #228, opened 2021-10-06
More service worker upgrade robustness and a playwright test

aomarks:
- Adds a **playwright test** to test service worker loading and upgrading in various scenarios.

- **Include the service worker version hash in the service worker scope**. Previously, we always registered the service worker at `<sandboxBaseUrl>/playground-projects`. Now, we register the service worker at `<sandboxBaseUrl>/__playground_swfs_<serviceWorkerHash>`.

  The service worker scope configures which URL paths the service worker will intercept. You can't have two service workers running on the same scope at the same time -- the newer one will replace the older one (with a somewhat complicated upgrade process).

  So by creating a unique scope for each version of the service worker, we bypass most of the issues relating to upgrade. Upgrades are now very unlikely to get stuck for any reason, because there shouldn't be any contention for its scope with other versions.

- **Rolls back the `unregister` change** from https://github.com/google/playground-elements/pull/218. Now that we have a unique scope for each version, there is no need to unregister.

- Makes service worker registration a little more robust, based on learning a little more about the lifecycle, and add some additional logging.

### DIFF
--- a/src/playground-service-worker-proxy.ts
+++ b/src/playground-service-worker-proxy.ts
@@ -71,12 +71,6 @@ import {
         // if we never end up getting a new version here. That would indicate
         // that the SW and project are out of date *on the server*.
         registration.update();
-        // There seems to be a bug in Chrome where the previous version of a
-        // service worker sometimes gets stuck and never shuts down when an
-        // updating is pending, even across a full browser restart. Calling
-        // unregister here seems to reliably force the previous version to shut
-        // down.
-        registration.unregister();
       }
     }
   );
--- a/src/playground-service-worker-proxy.ts
+++ b/src/playground-service-worker-proxy.ts
@@ -52,67 +52,122 @@ import {
     {scope}
   );
 
-  window.addEventListener(
-    'message',
-    (event: MessageEvent<PlaygroundMessage>) => {
-      if (event.data.type === UPDATE_SERVICE_WORKER) {
-        // When the project handshakes with the service worker, it may notice a
-        // version mismatch, in which case it will send this message to request
-        // the service worker update.
-        //
-        // Calling update triggers a fetch of the service worker. If the scripts
-        // are byte-different, then it will be installed + activated, and we'll
-        // receive the "updatefound" even there. Note this is exactly the normal
-        // Service Worker update lifecycle, it's just that calling update()
-        // forces it to happen immediately, as opposed to at some unknown
-        // opportunistic time in the future.
-        //
-        // TODO(aomarks) It would be good to show an error to the user somehow
-        // if we never end up getting a new version here. That would indicate
-        // that the SW and project are out of date *on the server*.
-        registration.update();
-      }
+  /** https://www.w3.org/TR/service-workers/#get-newest-worker-algorithm */
+  const getNewestWorker = () =>
+    registration.installing ?? registration.waiting ?? registration.active;
+
+  /** The service worker we most recently connected to. */
+  let newest: ServiceWorker | null = null;
+
+  /**
+   * Resolve when the given service worker reaches its activated state.
+   */
+  const activated = async (sw: ServiceWorker): Promise<void> => {
+    if (sw.state === 'activated') {
+      return;
     }
-  );
+    return new Promise<void>((resolve) => {
+      const aborter = new AbortController();
+      sw.addEventListener(
+        'statechange',
+        () => {
+          if (sw.state === 'activated') {
+            resolve();
+            aborter.abort();
+          }
+        },
+        {signal: aborter.signal}
+      );
+    });
+  };
 
-  const connect = (sw: ServiceWorker) => {
+  const connectToNewest = async () => {
+    const sw = getNewestWorker();
+    newest = sw;
+    if (sw === null) {
+      // Can't find a way to reproduce this scenario, possibly can't happen.
+      console.error('No playground service worker found.');
+      return;
+    }
+
+    // Don't connect to the project until the service worker is activated.
+    // Otherwise the project might try to load a preview before the service
+    // worker is actively controlling the preview's URL space.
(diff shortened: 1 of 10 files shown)


src/test/playwright/service-worker.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+  };
+  let expectedVersion: string;
+
+  const setServiceWorkerAndExpectedVersion = (version: string) => {
+    setServiceWorkerVersion(version);
+    setExpectedVersion(version);
+  };
+
+  /**
+   * Return a proimse that resolves the next time the service worker is served.
```
AndrewJakubowicz, line 76: typo: "promise"

src/test/playwright/service-worker.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+          throw new Error('Expected string body');
+        }
+        if (ctx.path.endsWith('playground-service-worker.js')) {
+          body = swCode
+            .replace(originalSwVersion, swVersion)
+            // Extreme hack! We need some way to check which service worker
+            // served a response. Prepend the version into all responses.
+            .replace(/new Response\(/g, `new Response("${swVersion} "+`);
+          serviceWorkerServedDeferred.resolve();
+          serviceWorkerServedDeferred = new Deferred();
```
AndrewJakubowicz, line 121: This is re-assigned at the start of each test as well. Are both needed?

src/test/playwright/service-worker.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+  ): Promise<void> => {
+    const text = await getIframeBodyText(page);
+    expect(text).toEqual(`${version} index.html`);
+  };
+
+  const clickReloadButton = async (page: Page): Promise<void> => {
+    await page.click('#reload-button');
+  };
+
+  const expectConsoleLog = (page: Page, expect: string): Promise<void> =>
```
AndrewJakubowicz, line 182: Does this test have a fail state? I.e. if there is a console message that doesn't match -> reject.

I'm assuming the fail state results in a timeout?

AndrewJakubowicz (approved):
LGTM!

The playwright setup is really nice.

src/playground-service-worker-proxy.ts
```diff
@@ -52,73 +52,122 @@ import {
...
+      console.error('No playground service worker found.');
+      return;
+    }
 
+    // Don't connect to the project until the service worker is activated.
+    // Otherwise the project might try to load a preview before the service
+    // worker is actively controlling the preview's URL space.
+    await activated(sw);
+    if (sw !== newest) {
+      // A new service worker appeared while we were waiting.
```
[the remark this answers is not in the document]
aomarks: It would have to be something very strange where version 0 started loading, and then a new version 1 was published to the server and the registration updated (maybe via another tab, or directly by the user in dev tools) somehow before version 0 finished installing.

Very good chance this would never ever happen in practice.

src/playground-service-worker-proxy.ts
```diff
@@ -52,73 +52,122 @@ import {
...
+    if (registration.installing !== newest) {
+      connectToNewest();
     }
   });
 
+  // A message from the service worker.
   navigator.serviceWorker.addEventListener(
     'message',
     (event: MessageEvent<PlaygroundMessage>) => {
-      if (event.data.type === MISSING_FILE_API && registration.active) {
```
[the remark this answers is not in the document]
aomarks: It's replaced by the fact that connectToNewest() now handles the null case itself.

src/test/playwright/service-worker.spec.ts
aomarks: Fixed

src/test/playwright/service-worker.spec.ts
aomarks: It allows waiting for the service worker to be served multiple times within the same test. Granted we don't actually do that currently.

src/test/playwright/service-worker.spec.ts
aomarks: Yeah it just relies on the timeout to fail. There are various other things in the console such as the "lit is in dev mode!" message, so we can't just wait for the next message that appears (plus even if we did, we'd still have to rely on a timeout to catch the case where it never logs anything).

src/playground-service-worker-proxy.ts
```diff
@@ -52,73 +52,122 @@ import {
...
+        {signal: aborter.signal}
+      );
+    });
+  };
 
-  const connect = (sw: ServiceWorker) => {
-    const {port1, port2} = new MessageChannel();
+  const connectToNewest = async () => {
+    const sw = getNewestWorker();
+    newest = sw;
```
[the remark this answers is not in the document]
aomarks: Good question! I originally factored it the way you suggested, but we need to account for the `UPDATE_SERVICE_WORKER` case at the bottom of this file. In that case, we need to force a reconnection even though we were previously connected (for when the service worker goes to sleep, then wakes up with amnesia).

Since now I think it actually is kinda unclear, I refactored it as described. There's now a force parameter which hopefully makes this a little more self describing than before.

I also renamed `newest` to `connected` because I think that makes a little more sense. And moved a variable down a little.

aomarks:
The playwright test is just too flakey to run in CI. I've filed https://github.com/google/playground-elements/issues/229 and will look into it further, but I feel fairly confident that these changes are an improvement, so I'd like to land them and investigate the tests later.

aomarks:
> The playwright test is just too flakey to run in CI. I've disabled them for now, filed #229 and will look into it further, but I feel fairly confident that these changes are an improvement, so I'd like to land them and investigate the tests as a followup.

Update: since it's mostly webkit that fails, I'll keep chrome running in CI.

PR #236, opened 2021-10-16
Fix broken deleting and renaming

aomarks:
Between MWC v0.25.1 and v0.25.2, when clicking on an `<mwc-icon-button>`, the event target changed from the `<mwc-icon-button>` itself, to the `<svg>` within it. We had some code that assumed the old target behavior, which broke the delete/rename context menu.


AndrewJakubowicz (approved):
LGTM!

Is the package lock update to keep deps up to date?

PR #239, opened 2021-11-11
[Feature] add property `file` to `playground-preview` 

jholt1:


aomarks:
Thanks for the PR! Feature makes sense to me. I'm on vacation currently so I'll take a closer look around the 15th.

aomarks:
Sorry for the very slow response here. We've actually just added very similar functionality via https://github.com/google/playground-elements/pull/249. The main difference from this PR is that the attribute is `html-file` instead of `file`.

It has just been released in playground-elements@0.14.8 -- hope this addresses your issue.

Thanks again for the PR!

PR #241, opened 2021-11-12
chore: Fix typo on importmap examples and reference to latest lit version

Matsuuu:
Heya, Trying to get back into speed with contributing. Set up some demos and noticed that the import maps demos had a typo making it not build

- Fix import map closing quotes
- Make reference in imports to lit 2.0.2


aomarks (approved):
Thanks!

PR #243, opened 2021-11-29
Code completions using Codemirror show hint

Matsuuu:
Heya! This is a long time coming, isn't it? 😄 

It's quite a large PR since I couldn't really cut this into smaller bits without sacrificing the bigger picture.

## What is this PR about?

This PR is about adding a "Code completion" functionality to the playground elements. I'd be glad to do an introduction through the huge code changes if need be. There's quite a lot that's happened. Some of the highlights:

- [Playground styles to apply by theme](https://github.com/google/playground-elements/pull/243/files#diff-1fc8310580b76754326514d9c01cc9731ffeafe9a1ef20c5e5fb496532ae4a41R375)
- [Code editor takes a set of completions in a 'dumb' way, and displays them and bubbled events upwards](https://github.com/google/playground-elements/pull/243/files#diff-bb71b6ede2a4fde7fe6e4bf1f7b4631936108f42c47e5c3fc7bc495608a2f29cL4)
- [File editor acts as a mediator between code editor and project components to pass completion info](https://github.com/google/playground-elements/pull/243/files#diff-f2da1a9d31c8bd90dcafb5d582d1a5d44b9bb93ec81b571e0eb12ce7c9c20cc7R12)
- [A new typescript lang service file that handles the completions and related actions](https://github.com/google/playground-elements/pull/243/files#diff-d02f028f8d5fd47fc6b423e658f6ef4e8365b7200b45184a58e0f39140d5aa5aR1)

## Added things

- [The show-hint addon to codemirror](https://github.com/google/playground-elements/pull/243/files#diff-c8156f10484ad01002c824bca5f1ecf0f08e622794e6edeb7b185a76e2d0bc43R1)
- [Fuse.js library for fussy matching and scoring completion items](https://github.com/google/playground-elements/pull/243/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R93)
- [The answer to life, universe and everything](https://www.youtube.com/watch?v=aboZctrHfK8)

Feel free to reach out to me on slack if you need some more thorough explanation or conversation


aomarks (changes requested):
This is really amazing work! Tried it out and it's working great!

My main comment is a thought about a slightly different way to handle fetching the active completion details that might let us simplify the API.

aomarks:
(Also sorry about the long review delay!)

src/typescript-worker/language-service-context.ts
```diff
@@ -85,6 +85,17 @@ class WorkerLanguageServiceHost implements ts.LanguageServiceHost {
     this._removeDeletedFiles(files);
   }
 
+  updateFile(filename: string, fileContent: string) {
```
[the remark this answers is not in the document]
Matsuuu: Yeah this was totally unnecessary. I think this was from the prototyping phase where I just needed something working. Will refactor it in the call-site

src/typescript-worker/completions.ts
```diff
@@ -0,0 +1,165 @@
...
+  const languageServiceHost = workerContext.languageServiceContext.serviceHost;
+  const searchWordIsPeriod = tokenUnderCursor.string === '.';
+
+  const options = {} as GetCompletionsAtPositionOptions;
+  if (searchWordIsPeriod) {
+    options.triggerCharacter = '.';
+  }
+
+  // Update language service status so that the file is up to date
+  languageServiceHost.updateFile(filename, fileContent);
```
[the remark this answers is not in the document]
Matsuuu: There always is the _potential_ that some kind of a bug could arise from this, but As it is now, (especially with fixing the call to use the `updateFileContentIfNeeded` -function), I don't see it causing too much trouble in the present. We should however keep it in mind, that it might cause something down the line.

I'll add a TODO to this code line, to remind ourselves about it

src/typescript-worker/completions.ts
```diff
@@ -0,0 +1,165 @@
...
+  const languageService = workerContext.languageServiceContext.service;
+  const languageServiceHost = workerContext.languageServiceContext.serviceHost;
+  const searchWordIsPeriod = tokenUnderCursor.string === '.';
+
+  const options = {} as GetCompletionsAtPositionOptions;
+  if (searchWordIsPeriod) {
+    options.triggerCharacter = '.';
+  }
+
+  // Update language service status so that the file is up to date
```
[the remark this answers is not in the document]
Matsuuu: Yeah, that's exactly the reason. If we are relying on the build step and it's state, we get rubbish completion, since it might lose the context for the completion. This way we can ensure that we have the actual current file context to give TS LS

src/typescript-worker/completions.ts
```diff
@@ -0,0 +1,165 @@
...
+  // This list of completions is quite vast, and therefore we will need to do some extra sorting
+  // and filtering on it before sending it back to the browser.
+  const completions = languageService.getCompletionsAtPosition(
+    filename,
+    cursorIndex,
+    options
+  );
+
+  if (searchWordIsPeriod) {
+    // On period, just return the completions without fuzzy finding since there's no need for it
```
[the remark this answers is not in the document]
Matsuuu: image
So the reason for this is that we don't have a "keyword" to fuzzy find with.

e.g. if we have written `console.l`, we can match words starting with "l".

I'll clear it out in the comment a bit more.

src/typescript-worker/completions.ts
```diff
@@ -0,0 +1,165 @@
...
+    const editorCompletions =
+      completions?.entries
+        .sort((a, b) => parseInt(a.sortText) - parseInt(b.sortText))
```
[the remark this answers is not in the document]
Matsuuu: The sorttext provided by TS LS is a helper for sorting the results. The LS however already returns the list sorted so this sorting function is unnecessary and removed now.

The reason we don't use this sorting function to sort our completions is, that in most cases, it gives the same value for a ton of completion items, even though some of them are better matches than others.

src/playground-project.ts
```diff
@@ -546,6 +563,56 @@ export class PlaygroundProject extends LitElement {
     this.dispatchEvent(new CustomEvent('compileDone'));
   }
 
+  /**
+   * Query the language service for completion options on
+   * token under cursor in code-editor
+   * */
+  async getCompletions(
```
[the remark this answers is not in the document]
Matsuuu: Renamed the getCompletions to updateCompletions for clarity

Matsuuu:
Allrighty. I've gone through your comments, and reworked some of the parts. I'll do some refactoring / beautifying tomorrow, but if you want to check it out in it's current state, I've pushed it already.

Matsuuu:
Allright, now it should be good to go. I've gone through the review fix recommendations, and acted upon them. On top of that as I was looking into reworking the details system to work in a promise based manner, I realized that we are doing way too many round trips to the TSLS Worker. I reduced this by re-using the completions from previous completion queries, if the situation allowed it, just re-filtering it through.

https://github.com/google/playground-elements/pull/243/files#diff-8f05853e3ce4da716aed3677ad38b34cf6423aaf403286a5a8ae458b97a85f5cR589-R642


This makes it blazingly much faster to get the completions, as the TSLS would just provide the same collection of completions, wether we wrote `document.que` or `document.queryS`.

---

On top of that, I've implemented a proxy-pattern -ish way of handling the details in our completions, in which the details are fetched in case they are missing, and stored as a promise into the completions. This means that if the completion is missing, we will be able to fetch it by calling a function inside the completion, and the promise containing the details will then be assigned to the details field of the completions 

https://github.com/google/playground-elements/pull/243/files#diff-8f05853e3ce4da716aed3677ad38b34cf6423aaf403286a5a8ae458b97a85f5cR666-R675

This makes it so that we can go through completions, fetching details as we need them, but not having to fetch already fetched details when navigating back to already visited completions.

---

This whole overhaul reduced a LOT of communication between the project and the TS worker, making the completion a lot smoother and nicer to use.

I hope it's in a good enough shape to go live :) @aomarks

src/playground-code-editor.ts
```diff
@@ -384,8 +407,29 @@ export class PlaygroundCodeEditor extends LitElement {
...
+        const isInputEvent = changeObject.origin === EditorChangeOrigin.INPUT;
+        const isCompletingCompletions =
+          (tokenUnderCursor.length > 1 || previousToken.string === '.') &&
+          isInputEvent;
+        const changeWasCodeCompletion =
+          changeObject.origin === EditorChangeOrigin.COMPLETE;
+        this.dispatchEvent(
+          new CustomEvent('change', {
+            detail: {
+              changeObject,
```
aomarks, line 427: Doesn't seem like the `changeObject` is used -- can it be removed?

aomarks (changes requested):
Looking good! Nice optimization for the refinement case. I have some thoughts on tightening up the event API a little further. Probably will have another round after this too with a few more comments, but looking good!

Sorry again for the slow review...

src/playground-code-editor.ts
```diff
@@ -384,8 +407,29 @@ export class PlaygroundCodeEditor extends LitElement {
...
+        const tokenUnderCursor = this.tokenUnderCursor.string.trim();
+        // To help reduce round trips to a language service or a completion provider, we
+        // are providing a flag if the completion is building on top of the earlier recommendations.
+        // If the flag is true, the completion system can just filter the already stored
+        // collection of completions again with the more precise input.
+        // On deletion events, we want to query the LS again, since we might be in a new context after
+        // removing characters from our code.
+        const isInputEvent = changeObject.origin === EditorChangeOrigin.INPUT;
+        const isCompletingCompletions =
+          (tokenUnderCursor.length > 1 || previousToken.string === '.') &&
```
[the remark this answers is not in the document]
Matsuuu: Could we lets the user provide a "completionTrigger" character or characters? So it would trigger on those, or when 2+ chars are input? And we could default the default trigger to . ?

src/playground-code-editor.ts
```diff
@@ -23,18 +40,10 @@ interface TypedMap<T> extends Map<keyof T, unknown> {
...
-  end: number;
-  /** Code string under the cursor. */
-  string: string;
-}
-
-interface EditorPosition {
-  ch: number;
-  line: number;
+export interface CodeEditorHint extends Hint {
+  details: undefined | Promise<EditorCompletionDetails>;
```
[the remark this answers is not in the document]
Matsuuu: Oh yeah a getter would do here. I was sure there was a better way. Will update

src/playground-code-editor.ts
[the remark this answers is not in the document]
Matsuuu: I've added a function that currently just checks if the type of the file is "ts" and doesn't ask for completions if it isn't. We can refine it down the line

Matsuuu:
Nudging this. I know yall have been busy with the new Lit releases but just in case :)

src/playground-project.ts
```diff
@@ -546,6 +563,88 @@ export class PlaygroundProject extends LitElement {
...
+      );
+    }
+    // We want to pre-fetch the first completion item, if it's present
+    // so that when the data gets to the code-editor, the detail hopefully
+    // is already loaded.
+    this._completions[0]?.details;
+    this.dispatchEvent(new CustomEvent('completionsChanged'));
+  }
+
+  async getCompletionDetails(
```
aomarks, line 632: Can `getCompletionDetails` be private? Looks like it's only referenced from within this class.

src/playground-code-editor.ts
```diff
@@ -189,6 +201,13 @@ export class PlaygroundCodeEditor extends LitElement {
   @property({attribute: false})
   diagnostics?: Array<Diagnostic>;
 
+  @property({attribute: false})
+  completions?: EditorCompletion[];
+
```
aomarks, line 206: I think we should add a `showCompletions` setting to `code-editor` and `file-editor`, so that completions can be turned on and off.

aomarks (changes requested):
Happy new year! Apologies again for the very long review delay.

I took a fresh look at this, and the main thing that occurred to me that there is an assumption about there being one set of completions per project that we should address, and in doing so we might also be able to even simplify the API quite a bit (comment in `playground-project.ts`).

src/playground-ide.ts
```diff
@@ -77,7 +77,6 @@ export class PlaygroundIde extends LitElement {
       height: 100%;
       flex: 1;
       min-width: 100px;
-      overflow: hidden;
```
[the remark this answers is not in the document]
Matsuuu: The removal was so that completion items would show even if the list overflew the editor

Matsuuu:
Heya @aomarks !

Obligatory "I haven't got a review all year" joke. 😁

Great comments and I agree in a lot of these points. I'll look into these with thought in the coming days!

Thanks again for the great review

src/playground-code-editor.ts
```diff
@@ -384,17 +406,236 @@ export class PlaygroundCodeEditor extends LitElement {
...
+    // On deletion events, we want to query the LS again, since we might be in a new context after
+    // removing characters from our code.
+    const isInputEvent = changeObject.origin === EditorChangeOrigin.INPUT;
+    const isRefinement =
+      (tokenUnderCursor.length > 1 || previousToken.string === '.') &&
+      isInputEvent;
+    const changeWasCodeCompletion =
+      changeObject.origin === EditorChangeOrigin.COMPLETE;
+
+    if (tokenUnderCursorAsString.length <= 0) return;
```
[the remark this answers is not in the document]
Matsuuu: That's pretty much not a case that can happen.

This is due to the fact that if the change was a completion, the completion item will still be under the cursor, since the completion item won't have a space or a line break in it.

src/playground-code-editor.ts
Matsuuu: Should we have completions enabled by default, what do you think? If so, the attribute would need to be reversed due to it being a boolean attribute. 

In the case that we enable completions by default, I suggest the attribute would be something like `no-completions`.

src/playground-project.ts
```diff
@@ -170,6 +179,14 @@ export class PlaygroundProject extends LitElement {
     return this._build?.diagnostics;
   }
 
+  get completions(): EditorCompletion[] | undefined {
```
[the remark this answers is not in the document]
bennypowers: > What if the request-completions event included a callback instead?

Sounds like a good case for context API community protocol https://github.com/lit/lit/pull/1955/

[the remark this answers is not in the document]
Matsuuu: Yeah this was a great idea and it wait actually really painless to implement.

src/playground-file-editor.ts
```diff
@@ -4,21 +4,22 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {html, css, PropertyValues} from 'lit';
-import {customElement, property, query} from 'lit/decorators.js';
-import {live} from 'lit/directives/live.js';
+import { html, css, PropertyValues } from 'lit';
```
aomarks, line 7: Looks like there was a lot of auto re-formatting in the latest commits. Mind re-running prettier?

Matsuuu: Ahh yeah sorry, done

src/playground-code-editor.ts
```diff
@@ -447,7 +451,14 @@ export class PlaygroundCodeEditor extends LitElement {
             new CustomEvent('request-completions', {
                 detail: {
                     isRefinement,
-                } as CodeEditorChangeData,
+                    fileName: this.fileName,
+                    fileContent: this.value,
+                    tokenUnderCursor,
+                    cursorIndex: this.cursorIndex,
+                    provideCompletions: (completions: EditorCompletion[]) => {
+                        this.completions = completions;
```
aomarks: I think a race could happen here, where e.g. completion request 1 and 2 are sent, but the answers are received in the reverse order. Once we've sent request 2, we should ignore any response for request 1.

One way to do this would be something like:

```ts
private _currentCompletionRequestId = 0;
...
const id = ++this._currentCompletionRequestId;
const provideCompletions = (completions: EditorCompletion[]) => {
  if (id === this._currentCompletionRequestId) {
    this.completions = completions;
  }
};
```

src/playground-project.ts
```diff
@@ -801,9 +877,9 @@ const typeEnumToMimeType = (type?: string) => {
   return undefined;
 };
 
-/**
+/*
```
aomarks, line 880: Another spot where some formatting got changed

src/playground-code-editor.ts
```diff
@@ -455,14 +458,25 @@ export class PlaygroundCodeEditor extends LitElement {
...
 
+  private _onCompletionsProvided(id: number, completions: EditorCompletion[]) {
+    // To prevent race conditioning, check that the completions provided
+    // are from the latest completions request.
+    if (id !== this._currentCompletionRequestId) {
+      return;
+    }
+
+    // Reset the counter on a successful completion
+    this._currentCompletionRequestId = 0;
```
aomarks, line 476: I don't think this should be reset, since that could cause the same kind of race condition. For example, this sequence:

request A
  current incremented to 1
request B
  current incremented to 2
response B
  B accepted
  current reset to 0
request C
  current incremented to 1
response A
  A accepted

Matsuuu: Ahh you're right. Somehow it felt like a good idea yesterday. Reverted now

src/playground-code-editor.ts
```diff
@@ -384,17 +411,252 @@ export class PlaygroundCodeEditor extends LitElement {
...
+
+    const hints = {
+      from: {line: lineNumber, ch: token.start} as Position,
+      to: {line: lineNumber, ch: token.end} as Position,
+      list: hintList,
+    } as Hints;
+
+    CodeMirror.on(
+      hints,
+      'select',
```
aomarks, line 505: Are these event listeners going to be garbage collected correctly? Looks like we're creating a new one every time the completions menu pops up? Is this delegating to `addEventListener` under the hood, or is it a bespoke codemirror event system?

Related: if we were writing our own completions UI, it would probably be a single container element, with a single event listener, which just gets updated and shown/hidden. Is codemirror's implementation creating a new element from scratch every time?

src/playground-code-editor.ts
```diff
@@ -384,17 +411,252 @@ export class PlaygroundCodeEditor extends LitElement {
...
+  }
+
   focus() {
     this._codemirrorEditable?.focus();
   }
 
+  private _hintFunction(
+    cm: Editor
+  ): Hints | null | undefined | PromiseLike<Hints | null | undefined> {
+    const cursorPosition = cm.getCursor('start');
```
aomarks, line 491: Isn't it possible for the cursor to have moved between requesting the completions and the completions being provided? Would it then be possible for the user to select a completion which would incorrectly get written to the new cursor location, instead of the original location?

src/playground-code-editor.ts
```diff
@@ -384,17 +411,252 @@ export class PlaygroundCodeEditor extends LitElement {
...
+    // Currently we are only supporting code completion for TS. Change
+    // this in a case that we start to support it for other languages too.
+    return this.type === 'ts';
+  }
+
   focus() {
     this._codemirrorEditable?.focus();
   }
 
+  private _hintFunction(
```
aomarks, line 488: How about renaming this to `_completionsAsHints`, and inlining the current `_completionsAsHints` function into this one. I think that would be easier to read, since they're very closely related.

src/playground-code-editor.ts
```diff
@@ -384,17 +411,252 @@ export class PlaygroundCodeEditor extends LitElement {
...
+      hint.displayText,
+      completionData
+    );
+    // Render the actual completion item first
+    this._renderCompletionItem(objectName, element);
+
+    // And if we have the detail promise passed into this function,
+    // we want to asynchronously update the detail info into our completion
+    // item. We don't want to block the rendering, so we don't use await.
+    if (detail !== undefined) {
```
aomarks, line 549: Would we ever not have a detail promise?

src/playground-code-editor.ts
```diff
@@ -384,17 +411,252 @@ export class PlaygroundCodeEditor extends LitElement {
...
+    target: HTMLElement
+  ) {
+    render(
+      html`<span class="hint-object-name">${objectName}</span>
+        <span class="hint-object-details">${details.text}</span> `,
+      target
+    );
+  }
+
+  private _buildHintObjectName(
```
aomarks, line 585: Can you document this method a bit? Not 100% sure what's going on here without digging deeper. How is the result being used, how is the `<mark>` being used, why is there an issue with padding, etc.

src/playground-code-editor.ts
```diff
@@ -384,17 +411,252 @@ export class PlaygroundCodeEditor extends LitElement {
...
+    objectName: string | undefined,
+    completionData: EditorCompletion | undefined
+  ) {
+    const markedObjectName = objectName ?? '';
+    const matches = completionData?.matches ?? [];
+    let padding = 0;
+    if (matches.length <= 0) {
+      return html`${markedObjectName}`;
```
aomarks, line 593: No need to tag this with `html` I think, lit will render a plain string.

src/playground-code-editor.ts
```diff
@@ -189,6 +202,15 @@ export class PlaygroundCodeEditor extends LitElement {
   @property({attribute: false})
   diagnostics?: Array<Diagnostic>;
 
+  @property({attribute: false})
+  completions?: EditorCompletion[];
+
+  private _onSelectedChange?: () => void;
+
+  private _currentSelectionLabel = '';
```
aomarks, line 210:
```suggestion
  private _currentCompletionSelectionLabel = '';
```

src/playground-code-editor.ts
```diff
@@ -189,6 +202,15 @@ export class PlaygroundCodeEditor extends LitElement {
   @property({attribute: false})
   diagnostics?: Array<Diagnostic>;
 
+  @property({attribute: false})
+  completions?: EditorCompletion[];
+
+  private _onSelectedChange?: () => void;
```
aomarks, line 208:
```suggestion
  private _onCompletionSelectedChange?: () => void;
```

src/playground-code-editor.ts
aomarks: Hmm, yeah let's have them on by default. So I agree, a `no-completions` attribute / `noCompletions` property sounds right. It should be on `code-editor`, `file-editor`, and `ide`.

src/playground-project.ts
```diff
@@ -546,6 +555,75 @@ export class PlaygroundProject extends LitElement {
     this.dispatchEvent(new CustomEvent('compileDone'));
   }
 
+  async provideCompletions(changeData: CodeEditorChangeData) {
```
aomarks, line 558:
```suggestion
  async getCompletions(changeData: CodeEditorChangeData) {
```

src/playground-project.ts
```diff
@@ -803,7 +879,7 @@ const typeEnumToMimeType = (type?: string) => {
 
 /**
  * Validate an import map configuration (https://wicg.github.io/import-maps/).
- * Returns an array of errors. If empty, the import map is valid.
+ * Returns an array of errors.If empty, the import map is valid.
```
aomarks, line 882: Restore the space here

src/playground-project.ts
```diff
@@ -680,7 +757,6 @@ const fetchProjectConfig = async (
     alreadyFetchedConfigUrls
   );
 };
-
 /**
```
aomarks, line 760: Restore the newline above

src/playground-project.ts
```diff
@@ -639,7 +717,6 @@ export class PlaygroundProject extends LitElement {
     this.save();
   }
 }
-
 /**
```
aomarks, line 720: Restore the newline above

aomarks (changes requested):
It would also be great to add some tests if possible, and also update the API tables in the README with the new public methods/events/attributes.

src/playground-code-editor.ts
```diff
@@ -384,17 +411,252 @@ export class PlaygroundCodeEditor extends LitElement {
...
+        // then triggered when moving to another selection, removing the details
+        // text from the previously selected element.
+        this._onSelectedChange = () => this._renderHint(element, _data, hint);
+        this._currentSelectionLabel = hint.text;
+      });
+    }
+  }
+
+  private _renderCompletionItem(
+    objectName: DirectiveResult,
```
aomarks, line 567: Unsure about the use of `DirectiveResult` type here. I think maybe it could be `string|TemplateResult`?

src/playground-code-editor.ts
Matsuuu: The detail promise is passed into this function only for the item currently highlighted from the completion list.

I've added documentation regarding it above the check

src/playground-code-editor.ts
Matsuuu: Yeah this actually seems to be the case on a slower connection/machine.

What do you suggest we do for this case? We would need to store the coordinates from which the latest completion request was sent, and use them?

aomarks: Maybe we shouldn't display the completions at all if the cursor has moved?

src/playground-code-editor.ts
Matsuuu: AFAIK from my research it attaches a eventlistener with `addEventListener` to the `ul`  element containing all of the completion items. Then as it closes, it removes the whole element from DOM.

So there is only one eventlistener per action, no matter the amount of completion items listed.

So...

 > Is this delegating to addEventListener under the hood, or is it a bespoke codemirror event system?
 
 It is attaching a single eventlistner per action (e.g. dblclick, click, etc.) to the UL element wrapping the list
 
 >  Is codemirror's implementation creating a new element from scratch every time?
 
  Yeah. It creates it from scratch, appends `li` items inside it and as it's closed, it rids of it with parentNode.removeChild(hints)

Matsuuu:
Went through most of the cases. Will look into implementing some tests for this next week or so

src/playground-code-editor.ts
```diff
@@ -279,18 +290,19 @@ export class PlaygroundCodeEditor extends LitElement {
             break;
           case 'diagnostics':
             this._showDiagnostics();
+            this._showCompletions();
```
aomarks, line 293: Why do we need to show completions when diagnostics change?

Matsuuu: This was a typo from a merge. Fixed in local

src/playground-code-editor.ts
Matsuuu: Yup. Made it check that the cursor index hasn't changed since the request, and reject the completions if it has

Matsuuu:
Heya, Sorry for the slight delay on continuing this, had a small break from open source projects due to life being busy.

I wrote out some integration/unit tests for the completions and they test some basic cases of the completion functioning.

src/test/completions_test.ts
```diff
@@ -0,0 +1,280 @@
...
+      node = result as HTMLElement;
+    }
+    return node;
+  };
+  const waitForCompileDone = () =>
+    new Promise((resolve) => {
+      project?.addEventListener(
+        'compileDone',
+        async () => {
+          waitMilli(1000);
```
aomarks, line 177: Are these extra waits because there's no explicit signal we can listen to that we've finished rendering completion UI after each change? Maybe another approach could be to poll every N milliseconds until the DOM we expect has appeared, timing out after M seconds?

aomarks (commented):
> When there's time, I'd love to have some pointers on the "finishing touches" this needs to be ready for actual use :)

What kind of touches were you thinking of? I think it's pretty much ready!

Matsuuu:
> I think it's pretty much ready!

Awesome! I'll check if I could write the polling function for the dom elements for the tests and then try it out through the CI :)

Matsuuu:
Yeah had been super silly with the tests. Rewrote the waits and they seem to pass way more stabily.

I think its good to go now!

aomarks:
> I think its good to go now!

Looks like it needs one last auto-format?

aomarks:
Merged! Amazing work on this.

I'm planning to do do a pre-release shortly today, and then early next week I'll do a test run of integrating it into the lit.dev playground. Assuming all looks good, we'll do a final release then!

Matsuuu:
Sounds great! I can try to do some demoing tomorrow too to make sure everythings fine

bennypowers:
Congrats @Matsuuu and @aomarks I've been following this thread the whole way, heady stuff. Great job!

PR #249, opened 2022-01-25
feat: add htmlFile property to ide and preview

### LINKED ISSUE #247: Playground file not found with direct config
Ran into the following case where the key in the files map caused a `Playground file not found` even when `content` was set.

```js
const failure = document.createElement('playground-ide');

failure.config = {
    files: {
        "some/long/path/to/file.html": {
            content: `<!DOCTYPE html><body><h1>Hello World</h1></body>`,
            label: "index.html"
        }
    }
}

document.body.appendChild(failure);

const success = document.createElement('playground-ide');

success.config = {
    files: {
        "index.html": {
            content: `<!DOCTYPE html><body><h1>Hello World</h1></body>`,
            label: "index.html"
        }
    }
}

document.body.appendChild(success);
```

The `failure` playground result is `Playground file not found`.

Screenshot

Reproducible example at: https://github.com/jpoehnelt/playground-elements-reproducible

jpoehnelt:
Add option for specifying the html file to render in the preview. 

cc #239
closes #247

### DIFF
--- a/src/playground-ide.ts
+++ b/src/playground-ide.ts
@@ -255,6 +255,12 @@ export class PlaygroundIde extends LitElement {
   pragmas: 'on' | 'off' | 'off-visible' = 'on';
 
   /**
+   * The HTML file used in the preview.
+   */
+  @property()
+  htmlFile = 'index.html';
+
+  /**
    * Indicates whether the user has modified, added, or removed any project
    * files. Resets whenever a new project is loaded.
    */
@@ -321,6 +327,7 @@ export class PlaygroundIde extends LitElement {
                        preview-loading-indicator,
                        diagnostic-tooltip,
                        dialog"
+          .htmlFile=${this.htmlFile}
           .project=${projectId}
         ></playground-preview>
       </div>
--- a/src/playground-ide.ts
+++ b/src/playground-ide.ts
@@ -257,7 +257,7 @@ export class PlaygroundIde extends LitElement {
   /**
    * The HTML file used in the preview.
    */
-  @property()
+  @property({attribute: 'html-file'})
   htmlFile = 'index.html';
 
   /**
--- a/src/playground-preview.ts
+++ b/src/playground-preview.ts
@@ -98,7 +98,7 @@ export class PlaygroundPreview extends PlaygroundConnectedElement {
   /**
    * The HTML file used in the preview.
    */
-  @property()
+  @property({attribute: 'html-file'})
   htmlFile = 'index.html';
 
   /**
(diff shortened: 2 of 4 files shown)


src/playground-ide.ts
```diff
@@ -254,6 +254,12 @@ export class PlaygroundIde extends LitElement {
   @property()
   pragmas: 'on' | 'off' | 'off-visible' = 'on';
 
+  /**
+   * The HTML file used in the preview.
+   */
+  @property()
```
aomarks, line 260: We've been following the convention that attributes are snake-case, so we should add (also in the other spot):

```suggestion
  @property({attribute: 'html-file')
```

src/playground-preview.ts
```diff
@@ -95,6 +95,12 @@ export class PlaygroundPreview extends PlaygroundConnectedElement {
     }
   `;
 
+  /**
+   * The HTML file used in the preview.
+   */
+  @property()
```
aomarks, line 101:
```suggestion
  @property({attribute: 'html-file')
```

src/test/playground-ide_test.ts
```diff
@@ -649,4 +649,19 @@ suite('playground-ide', () => {
...
+    ide.htmlFile = 'src/index.html';
+    container.appendChild(ide);
+    ide.config = {
+      files: {
+        'src/index.html': {
+          content: 'Hello HTML',
+        },
+      },
+    };
+    await assertPreviewContains('Hello HTML');
```
aomarks, line 665:
```suggestion
    await assertPreviewContains('Hello HTML');
    
    ide.htmlFile = 'other.html';
    await assertPreviewContains('Different content');
```

src/test/playground-ide_test.ts
```diff
@@ -649,4 +649,19 @@ suite('playground-ide', () => {
...
+  test('uses custom htmlFile property', async () => {
+    const ide = document.createElement('playground-ide')!;
+    ide.sandboxBaseUrl = '/';
+    ide.htmlFile = 'src/index.html';
+    container.appendChild(ide);
+    ide.config = {
+      files: {
+        'src/index.html': {
+          content: 'Hello HTML',
+        },
```
aomarks, line 662: Let's also make sure that the preview updates if you dynamically change this property:

```suggestion
        },
      'other.html': {
        content: 'Different content',
      }
```

aomarks (changes requested):
Awesome, thank you for the pr!

aomarks (approved):

Issue #154, opened 2021-05-13
Undo history applies across files


aomarks:
If you make a change, then switch files, then ctrl-z, weird stuff happens, like the previous file's contents moving into the current file.

PR #258, opened 2022-02-09
Fix leaky history by isolating CodeMirror document instances by key

AndrewJakubowicz:
Fixes: https://github.com/google/playground-elements/issues/154

### Context

Adds a new property, `documentKey` which can be used to isolate history and CodeMirror document instances.

### Why

Currently setting `value` programmatically, or via a user edit in the codemirror instance results in `undo`/`redo` history being added to a single instance.

This becomes a pain point when modeling a file system on top of the `<playground-code-editor>`, because changing to a new file is interpreted by CodeMirror as the user inserting the new file value into a single global document instance. This results in bugs where it's possible to trigger a file change and then "undo" the file change value.

See video of file change followed by undo (can repro at https://lit.dev/playground/):

https://user-images.githubusercontent.com/15080861/153515054-a05a0d05-bf03-43bb-9282-064a14f92bfc.mov

### How

`documentKey` drives an internal CodeMirror document instance. Changing the `documentKey` will either create a new document instance or reuse a previous one. Each document instance has isolated history.

The `value` property and text typed by the user is still the source of truth, and the `documentKey` just associates which CodeMirror instance is being edited.

### Testing

This has been tested manually via configurator, as well as with many programmatic tests as I was understanding each behavior.

 - `playground-code-editor_test.ts` demonstrates programmatic changes to `value` and `documentKey`.
 - `playground-ide_test.ts` emulates direct edits to the codemirror instance, file swapping, and undo/redo actions.
 - Note that renaming a file preserves file history because the file object reference is unchanged.
 - Note that changing **only** the documentKey doesn't drive a `value` change. On contention between the value cached on the document instance and `value` property, the document instance is updated to match the `value`.

### Backwards compatibility

The `<playground-code-editor>` remains backwards compatible if `documentKey` is not used. Any of the elements using a virtual file system will no longer allow undoing into a previous file's contents.

### DIFF
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -183,16 +183,13 @@ export class PlaygroundCodeEditor extends LitElement {
   }
 
   /**
-   * Alternative to `value`, and takes precedence over `value`. Use `docCache`
-   * to pass a CodeMirror document instance directly. If `docCache` is set,
-   * `value` will no longer update the contents of the editor. The document
-   * instance should be used if history needs to be preserved.
-   *
-   * Whenever a new document instance is created, a `newdocinstance` event is
-   * fired containing a `codeMirrorDoc` property.
+   * Providing a `documentKey` creates a CodeMirror document instance which
+   * isolates history and text changes between documents.
    */
   @property({attribute: false})
-  docCache?: Doc;
+  documentKey?: object;
+
+  private readonly _privateDocCache = new WeakMap<object, Doc>();
 
   /**
    * The type of the file being edited, as represented by its usual file
@@ -280,28 +277,32 @@ export class PlaygroundCodeEditor extends LitElement {
       this._createView();
     } else {
       const changedTyped = changedProperties as TypedMap<
-        Omit<
-          PlaygroundCodeEditor,
-          keyof LitElement | 'render' | 'update' | 'createNewDocInstance'
-        >
+        Omit<PlaygroundCodeEditor, keyof LitElement | 'render' | 'update'>
       >;
       for (const prop of changedTyped.keys()) {
         switch (prop) {
-          case 'docCache':
-            if (this.docCache && cm.getDoc() !== this.docCache) {
-              this._valueChangingFromOutside = true;
-              cm.swapDoc(this.docCache);
-              this._valueChangingFromOutside = false;
+          case 'documentKey': {
+            if (!this.documentKey) {
+              break;
             }
+            let docInstance = this._privateDocCache.get(this.documentKey);
+            if (docInstance === undefined) {
+              docInstance = new CodeMirror.Doc(this.value ?? '');
+            }
+            this._privateDocCache.set(this.documentKey, docInstance);
+            this._valueChangingFromOutside = true;
+            cm.swapDoc(docInstance);
+            this._valueChangingFromOutside = false;
             break;
+          }
           case 'value':
-            // If there is a CodeMirror `docCache`, we prefer that instead of
-            // setting a value.
-            if (this.docCache) {
+            if (changedTyped.has('documentKey')) {
+              // We've already processed the `documentKey` and don't want to
+              // add history by applying the `value` again.
               break;
             }
             this._valueChangingFromOutside = true;
-            cm.swapDoc(this.createNewDocInstance(this.value ?? ''));
+            cm.setValue(this.value ?? '');
             this._valueChangingFromOutside = false;
             break;
           case 'lineNumbers':
@@ -338,18 +339,6 @@ export class PlaygroundCodeEditor extends LitElement {
       }
     }
     super.update(changedProperties);
-  }
-
-  /**
-   * Create a new CodeMirror Doc instance.
-   * @param text Contents to pre-populate the CodeMirror Doc instance.
-   */
-  createNewDocInstance(text: string): Doc {
-    const codeMirrorDoc = new CodeMirror.Doc(text);
-    const changeEvt = new Event('newdocinstance');
-    (changeEvt as Event & {codeMirrorDoc: Doc}).codeMirrorDoc = codeMirrorDoc;
-    this.dispatchEvent(changeEvt);
-    return codeMirrorDoc;
   }
 
   render() {
@@ -462,10 +451,7 @@ export class PlaygroundCodeEditor extends LitElement {
         this._applyHideAndFoldRegions();
         this._showDiagnostics();
       } else {
-        const changeEvt = new Event('change');
-        (changeEvt as Event & {codeMirrorDoc: Doc}).codeMirrorDoc =
-          _editorInstance.getDoc();
-        this.dispatchEvent(changeEvt);
+        this.dispatchEvent(new Event('change'));
         this._requestCompletionsIfNeeded(changeObject);
       }
     });
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -21,6 +21,7 @@ import playgroundStyles from './playground-styles.js';
 import './internal/overlay.js';
 import type {Diagnostic} from 'vscode-languageserver';
 import type {
+  Doc,
   Editor,
   EditorChange,
   Hint,
@@ -182,6 +183,18 @@ export class PlaygroundCodeEditor extends LitElement {
   }
 
   /**
+   * Alternative to `value`, and takes precedence over `value`. Use `docCache`
+   * to pass a CodeMirror document instance directly. If `docCache` is set,
+   * `value` will no longer update the contents of the editor. The document
+   * instance should be used if history needs to be preserved.
+   *
+   * Whenever a new document instance is created, a `newdocinstance` event is
+   * fired containing a `codeMirrorDoc` property.
+   */
+  @property({attribute: false})
+  docCache?: Doc;
+
+  /**
    * The type of the file being edited, as represented by its usual file
    * extension.
    */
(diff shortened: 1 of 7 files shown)


src/playground-code-editor.ts
```diff
@@ -181,6 +182,18 @@ export class PlaygroundCodeEditor extends LitElement {
...
+  /**
+   * Alternative to `value`, and takes precedence over `value`. Use `docCache`
+   * to pass a CodeMirror document instance directly. If `docCache` is set,
+   * `value` will no longer update the contents of the editor. The document
+   * instance should be used if history needs to be preserved.
+   *
+   * Whenever a new document instance is created, a `newdocinstance` event is
+   * fired containing a `codeMirrorDoc` property.
+   */
+  @property({attribute: false})
```
aomarks, line 194: What do you think about this slight variation, which shifts the responsibility for the actual codemirror document caching into the hidden details of this module?

- Add a property `@property({attribute: false} documentKey?: object`
- Add a property `private readonly _documentCache = new WeakMap<object, Doc>()`.
- When a new `documentKey` is seen, create a new CodeMirror doc, and add it to the weakmap, and swap to it.
- When a known `documentKey` is set, get it from the cache and swap to it.

Then in `file-editor`, we can just have a binding like:

```ts
<playground-code-editor .documentKey=${this._currentFile}>
```

So, the identity of the `_currentFile` object itself serves as the key, and `code-editor` handles all of the details. The `WeakMap` is used so that the `code-editor` doesn't become a source of memory leaks when deleting files, switching projects, etc.

I think some advantages of this could be:

- Slightly simpler API. Only the one new public property is needed.

- Preserves the characteristic of not exposing any CodeMirror implementation details to public API. I've been thinking this is a nice thing to preserve, in particular because it would make it really easy to have drop-in replacements for `<playground-code-editor>` (such as a Monaco implementation), without `<playground-file-editor>` needing to know the difference.

It could also be nice as part of this to change the behavior of `value` when there is no `documentKey`:

- If `documentKey` is `undefined`, then whenever `value` is set *from the outside*, always create a new CodeMirror doc and swap to it. I think this is a better default than what we have now, because it's worse to have incorrect undo history, than no undo history at all.

aomarks (changes requested):
Awesome! Excited to see this getting fixed. Great tests too.

I have a suggestion for a variation on the approach, details in the comment!

src/playground-code-editor.ts
```diff
@@ -267,11 +279,43 @@ export class PlaygroundCodeEditor extends LitElement {
...
+          if (docInstance === undefined) {
+            docInstance = new CodeMirror.Doc(this.value ?? '');
+          } else {
+            // A Map keeps track of insertion order. This will cause an
+            // insertion maintaining an LRU invariant.
+            this._privateDocCache.delete(this.documentKey);
+          }
+          this._privateDocCache.set(this.documentKey, docInstance);
+          if (this._privateDocCache.size > 50) {
+            // Map is bound to 100 elements. The first key returned by `keys`
```
AndrewJakubowicz, line 299: oops - docs out of date.

src/playground-code-editor.ts
```diff
@@ -182,17 +182,16 @@ export class PlaygroundCodeEditor extends LitElement {
...
+   * Will keep 50 document instances before exhibiting LRU cache behavior. On
+   * reaching max size, the least recently used document history will be lost.
    *
-   * Whenever a new document instance is created, a `newdocinstance` event is
-   * fired containing a `codeMirrorDoc` property.
+   * This prevents memory leaks and unbounded growth of the cache.
    */
-  @property({attribute: false})
-  docCache?: Doc;
+  private readonly _privateDocCache = new Map<string, Doc>();
```
aomarks, line 194: Sorry not sure if this was actually intended to be reviewed or not, but why not use a WeakMap here?

AndrewJakubowicz: I could not seem to get the file objects to cache hit against a WeakRef. Going to try the WeakRef again because now I have this working and a manual test via configurator works well, and other places have been cleaned up.

AndrewJakubowicz: SMH, it works. I must have had some of the previous implementation causing an issue.

src/playground-code-editor.ts
AndrewJakubowicz: Chose not to include the breaking change when no `documentKey` is present. I think it's reasonable to have a single history stack by default. And it's only when adding abstractions like a file system where the `documentKey` becomes useful.

src/playground-code-editor.ts
```diff
@@ -271,7 +289,33 @@ export class PlaygroundCodeEditor extends LitElement {
...
+          case 'documentKey': {
+            if (!this.documentKey) {
+              break;
+            }
+            let docInstance = this._privateDocCache.get(this.documentKey);
+            if (!docInstance) {
+              docInstance = new CodeMirror.Doc(this.value ?? '');
+              this._privateDocCache.set(this.documentKey, docInstance);
+            } else {
+              if (this.value && docInstance.getValue() !== this.value) {
```
aomarks, line 301:
```suggestion
            } else if (this.value && docInstance.getValue() !== this.value) {
```

src/playground-code-editor.ts
```diff
@@ -271,7 +289,33 @@ export class PlaygroundCodeEditor extends LitElement {
...
+              }
+            }
+            this._valueChangingFromOutside = true;
+            cm.swapDoc(docInstance);
+            this._valueChangingFromOutside = false;
+            break;
+          }
           case 'value':
+            if (changedTyped.has('documentKey')) {
+              // We've already processed the `documentKey` and don't want to
```
aomarks, line 315: How about the case where all of these are true in the same `update()`: (1) We didn't have a  `documentKey` before, (2) A `documentKey` is set, (3) A `value` is set. It seems like we might ignore the new `value` right now, because both the `documentKey` and the `value` switch cases `break` early in that case?

I wonder if it would help to hoist the `documentKey` handling above the `switch` statement all together, so that we can guarantee we always do the document swap before we look at any new `value`. Then the `value` case can probably just remain as it was? (To make type checking happy we could just have a no-op `case` for `documentKey` with a brief comment saying it was handled above).

aomarks (changes requested):
Awesome! Just a few thoughts about some edge cases that might need some minor tweaks.

src/playground-code-editor.ts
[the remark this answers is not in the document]
AndrewJakubowicz: Great point. Can also add a test.

src/playground-code-editor.ts
AndrewJakubowicz: If a value is on `hasChanged` it's already updated on the instance.
In the given case (and I think it's verified by tests), `documentKey` has changed so we will skip the `value` block. And in the `documentKey` block we instantiate a new document instance from `this.value`  and swap it into codemirror which populates the editor.

Another reason `documentKey` must handle `value` cases is that setting values add history.

However you've raised the case where `documentKey` is unset to undefined. That case would indeed break in both places, so I'll add a test and handling to it.

AndrewJakubowicz: Addressed by removing the `documentKey` undefined case, and instead have a fallback default documentKey associated with the playground editor instance.

aomarks: Cool!

Just a thought -- maybe a way of simplifying the implementation for handling the `undefined` case could be to have an internal-only cache key that represents "default" documents. That could give you less special casing.

Also I was wondering about what would happen in this sequence: (1) type something (2) set document key to something (3) type something (4) set document key to undefined (5) undo. Would (5) undo the thing typed in (1)? I would think it probably shouldn't, which implies that the "default document" should get reset each time the document key gets set to undefined.

aomarks: Oh sorry, I see I sent this just as you sent your latest commit :)

src/playground-code-editor.ts
```diff
@@ -271,7 +296,28 @@ export class PlaygroundCodeEditor extends LitElement {
       >;
       for (const prop of changedTyped.keys()) {
         switch (prop) {
+          case 'documentKey': {
+            const docKey = this.documentKey ?? this._fallbackDocInstance;
```
aomarks, line 300: Curious about that sequence I mentioned in the other comment. If we go (1) undefined (2) defined (3) undefined -- then does it make sense for there to be common history between (1) and (3)?

I think if we did not want it to be shared, the change could maybe just be this?:

```suggestion
            const docKey = this.documentKey ?? {};
```


I'm actually totally not sure what the behavior should be, curious what you think.

src/playground-code-editor.ts
```diff
@@ -181,6 +182,30 @@ export class PlaygroundCodeEditor extends LitElement {
...
+  /**
+   * Provide a `documentKey` to create a CodeMirror document instance which
+   * isolates history and value changes per `documentKey`.
+   *
+   * Use to keep edit history separate between files while reusing the same
+   * playground-code-editor instance.
+   */
+  @property({attribute: false})
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  documentKey?: object = this._fallbackDocInstance;
```
aomarks, line 199: `this._fallbackDocInstance` being the default here means that the special key can be read by somebody outside, and it also might be unintuitive that it would not be `undefined` when the user hasn't set it yet. Seems like we should probably keep the fallback key completely private?

aomarks (changes requested):

src/playground-code-editor.ts
AndrewJakubowicz: You are a genius. Your test case is great. I've codified it and played with both behaviors.

The behavior that "feels" right is what you suggested, that unsetting the `documentKey` should clear history. And I was mistakenly thinking that this was a breaking change. But it isn't. If the user never uses documentKey they get a single history.

Added a new test in newest commit for this case.

src/playground-code-editor.ts
AndrewJakubowicz: Your excellent suggestion in the other comment let me remove this idea. Also that is a great point that I hadn't considered.

AndrewJakubowicz:
Please enjoy your PTO! I'm very happy to wait till next week! Also your feedback is REALLY great!

aomarks (approved):
Fantastic! So glad this is fixed.

Issue #251, opened 2022-02-04
It appears only the first </script> (closing script) tag is unescaped in html files using playground-ide


jvoccia:
When I have a sandbox with code like the following:
https://codesandbox.io/s/naughty-kalam-i5e6h?file=/src/index.js

```html
<playground-ide editable-file-system line-numbers resizable>
  <script type="sample/html" filename="index.html"><!doctype html>
  <html>
      <head>
          <script src="script1.js">&lt;/script>
          <script src="script2.js">&lt;/script>
      </head>
      <body>
          <script src="./script3.js">&lt;/script>
      </body>
  </html></script>
  </playground-ide>
```

The resulting HTML in the IDE looks like:
```html
<!doctype html>
  <html>
      <head>
          <script src="script1.js"></script>
          <script src="script2.js">&lt;/script>
      </head>
      <body>
          <script src="./script3.js">&lt;/script>
      </body>
  </html>
```

Only the first script tag is "unescaped" the rest remain with &lt;/script>

AndrewJakubowicz:
Thank you for this report! I believe I have a fix implemented with https://github.com/google/playground-elements/pull/260

PR #260, opened 2022-02-11
Fix unescaping only a single closing script tag in playground-ide

AndrewJakubowicz:
Fixes https://github.com/google/playground-elements/issues/251

Previous logic only unescapes a single `&lt;`. Fixed by unescaping all instances of `&lt;/` such that we don't over-replace intentional escapes.

This might be a breaking change since previously any sample file with any `&lt;` would have the first one replaced with `<`. The new behavior should occur more closely to what we document, and limited to only `sample/html` files.

Added a unit test (that failed first).

Alternatively could make the unescaping logic even more aggressive and match for `&lt/script>`.

### DIFF
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -386,8 +386,12 @@ export class PlaygroundProject extends LitElement {
         continue;
       }
       const fileType = typeAttr.substring('sample/'.length);
-      // TODO (justinfagnani): better entity unescaping
-      let content = s.textContent!.replace('&lt;', '<');
+      let content = s.textContent ?? '';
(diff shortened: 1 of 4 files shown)


src/playground-project.ts
```diff
@@ -386,8 +386,12 @@ export class PlaygroundProject extends LitElement {
         continue;
       }
       const fileType = typeAttr.substring('sample/'.length);
-      // TODO (justinfagnani): better entity unescaping
```
aomarks, line 389: A prescient comment :grin:

aomarks (approved):
Great fix!

PR #262, opened 2022-02-14
Fix duplicate names on completion items

Matsuuu:
Fixes the duplication bug discussed here https://github.com/lit/lit.dev/pull/689

- Fix completion items showing the name multiple times in cases where
  there were multiple incdices matching

- Clean up and make completion item marking clearer code wise

### DIFF
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -683,34 +683,19 @@ export class PlaygroundCodeEditor extends LitElement {
       return markedObjectName;
     }
 
-    let padding = 0;
-    const markedObjectHTML = html`
-      ${matches.map((match) => {
-        const firstMatchingIndex = match.indices[0];
-        const start = firstMatchingIndex[0];
-        const end = firstMatchingIndex[1];
+    const firstMatch = matches[0];
 
-        const preMarkContent = markedObjectName?.substring(0, start + padding);
+    const firstMatchingIndex = firstMatch.indices[0];
+    const start = firstMatchingIndex[0];
+    const end = firstMatchingIndex[1];
 
-        const markedContent = markedObjectName?.substring(
-          start + padding,
-          end + padding + 1
-        );
+    const preMarkContent = markedObjectName?.substring(0, start);
+    const markedContent = markedObjectName?.substring(start, end + 1);
+    const postMarkedContent = markedObjectName?.substring(end + 1);
 
-        const postMarkedContent = markedObjectName?.substring(
-          end + padding + 1
-        );
-        const markedHTML = html`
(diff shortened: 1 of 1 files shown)


src/playground-code-editor.ts
```diff
@@ -677,36 +677,38 @@ export class PlaygroundCodeEditor extends LitElement {
...
       // In the situation, that none of the input matches with the
       // completion item suggestion, we exit early, leaving the objectName unmarked.
       return markedObjectName;
     }
 
+    let padding = 0;
     const markedObjectHTML = html`
-      ${matches.map(
-        (match) => html`
-          ${match.indices.map((ind) => {
```
AndrewJakubowicz, line 690: Ah I see, so the issue is that multiple indices would match. And fix is to take the first one.

src/playground-code-editor.ts
```diff
@@ -677,36 +677,38 @@ export class PlaygroundCodeEditor extends LitElement {
...
     const markedObjectName = objectName ?? '';
     const matches = completionData?.matches ?? [];
-    let padding = 0;
     if (matches.length <= 0) {
       // In the situation, that none of the input matches with the
       // completion item suggestion, we exit early, leaving the objectName unmarked.
       return markedObjectName;
     }
 
+    let padding = 0;
```
AndrewJakubowicz, line 686: I wasn't able to repro or understand when this is used. In examples that I tried, length of `matches` always seems to be `1`. Do we need the `padding` logic?

src/playground-code-editor.ts
```diff
@@ -677,36 +677,38 @@ export class PlaygroundCodeEditor extends LitElement {
...
+        );
+
+        const postMarkedContent = markedObjectName?.substring(
+          end + padding + 1
+        );
+        const markedHTML = html`
+          ${preMarkContent}<mark>${markedContent}</mark>${postMarkedContent}
+        `;
+        // As the matching is done in a fuzzy manner, we might have multiple matching
+        // indices in the completion word. In these situations, we need to pad out the
```
AndrewJakubowicz, line 707: Does this case occur anymore?

AndrewJakubowicz (commented):
This looks great and works great. Tested in configurator.

My only nit is that I don't quite understand the `padding` logic. Locally I added an error that throws if the `length` of `matches` is greater than 1, and couldn't get it to fire. In what cases are there multiple matches?

Thank you so much!

Matsuuu:
Ah yeah you're absolutely correct. The padding logic was a relic from a earlier implementation and was not needed anymore. Did the necessary changes and it's a lot clearer now

AndrewJakubowicz (approved):
Nice! Thank you!

aomarks (approved):

PR #263, opened 2022-02-14
Tweak CodeMirror hint styles

Matsuuu:
Modifying the code hint styles according to discussion in https://github.com/lit/lit.dev/pull/689

- Remove padding from hintlist
- Move removed padding to items
- Remove border-radius from hintlist
- Inherit font

Subjects up for discussion:

- Should we remove the box shadow from the completion items list? Does this affect visual clarity?
- What font weight should the completion items be?

### DIFF
--- a/playground-styles.css
+++ b/playground-styles.css
@@ -388,7 +388,7 @@ div.CodeMirror-dragcursors {
 
   background: var(--playground-code-background, white);
   font-size: var(--playground-code-font-size, 14px);
-  font-family: inherit;
+  font-family: var(--playground-code-font-family, monospace);
 
   max-height: 20em;
   width: 600px;
@@ -399,7 +399,6 @@ div.CodeMirror-dragcursors {
 .CodeMirror-hint {
   margin: 0;
   padding: 0 6px;
-  border-radius: 2px;
   white-space: pre;
   color: var(--playground-code-cursor-color, black);
   cursor: pointer;
@@ -419,7 +418,6 @@ div.CodeMirror-dragcursors {
 
 .CodeMirror-hint .hint-object-name {
   padding-right: 2em;
-  font-weight: bold;
   white-space: nowrap;
 }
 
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz (changes requested):
Thank you for making this PR!!!

I'd opt-out of having the `bold` option [here](https://github.com/google/playground-elements/pull/263/files#diff-1fc8310580b76754326514d9c01cc9731ffeafe9a1ef20c5e5fb496532ae4a41R422), as in configurator only `monospace` works with bold. The other font selections look the same.

And I'm happy to keep the shadow as well.

Otherwise this LGTM! Tested by randomly changing colors and things in the configurator.
Thank you so much!

Matsuuu:
Great catches here too! Did the necessary style tweaks

AndrewJakubowicz (changes requested):
Brilliant!

The only two last things that come to mind (that maybe we don't need to address):

1. Maybe we make a new CSS custom property for the [CodeMirror-hint-active selector](https://github.com/google/playground-elements/pull/263/files#diff-1fc8310580b76754326514d9c01cc9731ffeafe9a1ef20c5e5fb496532ae4a41R409-R412), instead of the filter to allow more control on the active color.

2. Responsiveness on narrow devices.

On Stackblitz there is some space between the hint and viewport: 
Screen Shot 2022-02-15 at 12 31 52 PM
Whilst currently on the preview lit.dev branch, the hint box touches the viewport edge:
Screen Shot 2022-02-15 at 12 33 25 PM

Matsuuu:
Heya,

I think the color customization is quite good as is as it is attached to the theming, and is inherited from the theme itself, instead of other custom properties.

As for the responsiveness, I added a small `min` css function to make the completion item list a bit more responsive on smaller screens. I don't  know if we should put too much effort into supporting mobile size screens due to them not being really the use case for a code editor in my opinion

AndrewJakubowicz (approved):
That makes sense. Changes LGTM.

I can make another pre-release of playground elements and try on lit.dev.
Thank you!

aomarks (approved):

PR #266, opened 2022-02-16
Close code completions menu on blur

Matsuuu:
Code completion items menu didn't close on blur, and it was because of a dev-time setting that was forgotten on.


AndrewJakubowicz (approved):
Looks great to me! Thanks!

Issue #267, opened 2022-02-17
`playground-hide` adds empty values on the undo history.


AndrewJakubowicz:
### Context

The logic for folding sets undo history that clears the file.
https://github.com/google/playground-elements/blob/51fa356543916631cc288a50d62b1ca388a20d1b/src/playground-code-editor.ts#L776

For repro, there is a test linking this issue number reproducing the bug.

This behavior can also be viewed on lit.dev, with the clock example. Navigate to [my-timer.ts](https://lit.dev/docs/), and change the file returning to `my-timer`. On undoing notice empty contents included in the undo stack.

AndrewJakubowicz:
Follow up from https://github.com/google/playground-elements/pull/281.

This PR introduces a behavior in CodeMirror where any readonly block will clear the history. Therefore the original issue no longer reproduces.

Next steps:

 - Should this edge case be documented?
 - Update test so this case is tested. Then if CodeMirror changes this behavior we'll know.
 - Maybe programmatic `.value` property updates should always clear history?

Matsuuu:
I would be going for number 2 or 3, or even both.

Having a test to track if this changes in the future, especially if we end up updating CodeMirror to the latest version would be great.

The situation where the value can be updated from the outside usually is one in which the content drastically changes, and in my opinion in these situations, it could be beneficial to clear the history.

I wonder if this will affect functionality with the playgrounds-ide though as the content is swapped when changing files, will the undo history just be erased when opened files are changed and changed back?

AndrewJakubowicz:
Agreed!

>I wonder if this will affect functionality with the playgrounds-ide though as the content is swapped when changing files, will the undo history just be erased when opened files are changed and changed back?

I manually tested the last case, and _think_ it works correctly because we:

1. Scope the history per virtual file.
2. Only set the fold/hide region on the initial `.value` setting of the virtual file (when the document instance is created).

Tested by updating configurator to have a folded region. Then edited beneath it - changed files a bit. Then checked that undo worked after swapping files.

However if later `value` is programmatically changed, it will clear the history (number 3).

Matsuuu:
I think if the cases you listed you tested work well, it would be 100% fine to clear the history on an external programmatic `value` change, as how I see it, the `value` change usually means "turning a new page" and changing the content/context, therefore it makes sense to ditch the history too

AndrewJakubowicz:
[...]

PR #268, opened 2022-02-17
Fix fold region regression with document instances

AndrewJakubowicz:
### Context

[The document instance commit](https://github.com/google/playground-elements/commit/29078343774ae023df7c6bfaf6630c51cf454c50#diff-bb71b6ede2a4fde7fe6e4bf1f7b4631936108f42c47e5c3fc7bc495608a2f29cR294-R308) caused a regression in code folding behavior.

The regression prevents code folding from occurring.

### Why?

Previously with a single instance, we'd attempt to code fold on every `change` event that is triggered on `setValue`. `swapDoc` doesn't trigger this code path and the folding logic never triggered.

### How was this fixed

This was fixed specifically for the document instance case. On document instance creation run the folding logic once to ensure folding occurs.

Tested with unit tests.

Also opened https://github.com/google/playground-elements/issues/267 to track an existing issue that continues to exist (although less severely) when running fold logic.

### DIFF
--- a/src/playground-code-editor.ts
+++ b/src/playground-code-editor.ts
@@ -294,9 +294,11 @@ export class PlaygroundCodeEditor extends LitElement {
           case 'documentKey': {
             const docKey = this.documentKey ?? {};
             let docInstance = this._docCache.get(docKey);
+            let createdNewDoc = false;
             if (!docInstance) {
               docInstance = new CodeMirror.Doc(this.value ?? '');
               this._docCache.set(docKey, docInstance);
+              createdNewDoc = true;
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz:
FYI @Matsuuu , blocking updating pre-release on this change since I also broke code folding 😢 

Thank you!

aomarks (approved):
Nice!

AndrewJakubowicz:
Thanks Al! Tests are failing because different browsers are giving me different zero width spaces.

Will resolve tests.

PR #277, opened 2022-03-01
feat: add `lhs` and `rhs` css parts

bennypowers:
Allows users more flexibility in styling playground-ide

### DIFF
--- a/src/playground-ide.ts
+++ b/src/playground-ide.ts
@@ -297,7 +297,7 @@ export class PlaygroundIde extends LitElement {
         <slot></slot>
       </playground-project>
 
-      <div id="lhs">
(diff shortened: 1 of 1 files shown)


justinfagnani:
I know it's been a year since you made this @bennypowers, but I just saw it scanning the repo. I don't know how @aomarks feels, but this seems like it make bake in a little too much about the internal structure into the public API. For instance, I could see a possible move to using CSS grid for laying out the tabs, editor, preview (and though it's not separate yet, the preview toolbar).

What kind of styling are you trying to do? Since the tab-bar, editor, and preview are all already parts, you can target most things. Are you trying to add gaps and such that could maybe be done with CSS variables?

PR #279, opened 2022-03-04
fix: use lowercase content-type header value

jpoehnelt:
Deployed to a server where `.ts` was being returned as `video/MP2T` instead of `video/mp2t` as required here. This particular header value is not case sensitive.

> The type, subtype, and parameter names are not case sensitive. For example, TEXT, Text, and TeXt are all equivalent. Parameter values are normally case sensitive, but certain parameters are interpreted to be case- insensitive, depending on the intended use. (For example, multipart boundaries are case-sensitive, but the "access- type" for message/External-body is not case-sensitive.)

https://www.w3.org/Protocols/rfc1341/4_Content-Type.html

### DIFF
--- a/src/playground-file-editor.ts
+++ b/src/playground-file-editor.ts
@@ -210,7 +210,7 @@ const mimeTypeToTypeEnum = (mimeType?: string) => {
   if (encodingSepIndex !== -1) {
     mimeType = mimeType.substring(0, encodingSepIndex);
   }
-  switch (mimeType.toLowerCase()) {
+  switch (mimeType) {
     // TypeScript: this is the mime-type returned by servers
     // .ts files aren't usually served to browsers, so they don't yet
     // have their own mime-type.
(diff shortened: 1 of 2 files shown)


aomarks (approved):
Thanks for the fix! I just added a CHANGELOG entry as well.

src/playground-file-editor.ts
```diff
@@ -210,7 +210,7 @@ const mimeTypeToTypeEnum = (mimeType?: string) => {
   if (encodingSepIndex !== -1) {
     mimeType = mimeType.substring(0, encodingSepIndex);
   }
-  switch (mimeType.toLowerCase()) {
```
aomarks: I think we should keep this normalization too -- because the `Content-Type` can also be set directly in the `project.json` file.

(Actually, do you know why this wasn't sufficient? There must be another place where we check the `Content-Type` that you were hitting?)

jpoehnelt: I believe this was sufficient. Just moved it earlier in case the `content-type` header gets used elsewhere.

PR #281, opened 2022-03-12
Make hidden blocks readonly to prevent accidental erasure

Matsuuu:
By default, the hidden blocks generated using `playground-hide` and codemirror aren't readonly, and therefore could be erased by the user just pressing the backspace while focusing on the editor.

To prevent this kind of erasure, we should have the hidden blocks as readonly.

This doesn't howevery prevent deleting the content with CTRL-A (select all) and then deleting it, but I'm not 100% sure if preventing that is possible with CodeMirror at all.


aomarks:
The effect now is that when you try to backspace into the hidden area, nothing happens at all, like the backspace key is broken.

I guess that's probably better, because not being able to see what you are deleting is potentially even more confusing.

I wonder if we could do something like show the region if you try and backspace into it, so that it's clear what's happening.

I'm ok with this change for now, though, seems like an improvement even if this is still not ideal.

Would you mind adding a `CHANGELOG.md` entry?

aomarks (approved):
Thanks for the pR!

Matsuuu:
> I wonder if we could do something like show the region if you try and backspace into it, so that it's clear what's happening.

yeah I think this could be good for consideration, but with the current setup, it would need quite a lot of diffing and noticing when commented blocks are broken/created, so for the time being, I think this is the best approach :)

aomarks (dismissed):
Looks like there is a newly failing test case. Since it's related to hidden regions, I suspect it is related to this change.

```
❌ playground-ide > code remains folded on value change and undo
      AssertionError: expected '…console.log(\'tomato\');' to equal ''
      + expected - actual
      
      -…console.log('tomato');
      
      at o.<anonymous> (test/playground-ide_test.js:[73](https://github.com/google/playground-elements/runs/5544028701?check_suite_focus=true#step:7:73)6:16)
```

https://github.com/google/playground-elements/runs/5544028701?check_suite_focus=true

Matsuuu:
The change only affected hidden parts, not folded parts. According to this comment, this test case is to reproduce a bug https://github.com/google/playground-elements/blob/main/src/test/playground-ide_test.ts#L1034-L1035

Could @AndrewJakubowicz clear out if this test was supposed to be failing in the current setup or if this change to the hidden blocks affected the fold code too, causing the test to break?

AndrewJakubowicz:
This was pretty interesting to investigate. This is my hypothesis for what is going on.

From CodeMirror docs [[link](https://codemirror.net/doc/manual.html#:~:text=A%20read%2Donly,the%20current%20implementation)]:

> Note: adding a read-only span currently clears the undo history of the editor, because existing undo events being partially nullified by read-only spans would corrupt the history (in the current implementation).

Even though the test uses `fold`, we always `hide` the comments that instruct code hiding or folding. Therefore in this `fold` test we end up with a hidden, read only, comment which clears the history as mentioned in the CodeMirror docs.

It's probably fine to clear the history if the `playground-code-editor` has its `value` property set from outside. Maybe that should be a default behavior?

To unblock this PR I recommend commenting out line 1087 (containing the final assert) of the failing test (instead of just the value) and I can update the issue.

```
    assert.equal(
      innerTextWithoutSpaces(
        codemirror?.shadowRoot?.querySelector<HTMLDivElement>('*')
      ),
      // This should be `EXPECTED_FOLDED`.
      // Issue: https://github.com/google/playground-elements/issues/267
      ''
    );
```

to

```
    // Issue: https://github.com/google/playground-elements/issues/267
    // assert.equal(
    //   innerTextWithoutSpaces(
    //     codemirror?.shadowRoot?.querySelector<HTMLDivElement>('*')
    //   ),
    //   EXPECTED_FOLDED
    // );
```

Matsuuu:
Thanks for the thorough investigation @AndrewJakubowicz . 

I modified the test case as instructed. Let's see the development on the fold/hide functionality and it's effect on the history in the other issue :)

CHANGELOG.md
```diff
@@ -17,7 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Fixed -->
 <!-- ### Removed -->
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Make hidden code blocks readonly to prevent accidental erasure
```
AndrewJakubowicz, line 22:
```suggestion

### Added

- Make hidden code blocks readonly to prevent accidental erasure.
```

AndrewJakubowicz (commented):
Thank you! I think the test will still fail - and it's ok to comment out the entire assertion.

src/test/playground-ide_test.ts
```diff
@@ -1088,9 +1088,7 @@ console.log('tomato');`;
       innerTextWithoutSpaces(
         codemirror?.shadowRoot?.querySelector<HTMLDivElement>('*')
       ),
-      // This should be `EXPECTED_FOLDED`.
-      // Issue: https://github.com/google/playground-elements/issues/267
-      ''
+      EXPECTED_FOLDED
     );
```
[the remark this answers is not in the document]
Matsuuu: Ahh silly me. Thought I had the test succeeding but must've had an old build. Sorry about this. Will fix it now

AndrewJakubowicz (approved):
Thank you!

AndrewJakubowicz:
Very strangely the CI checks aren't showing up for this PR. Going to try making a tiny change to refresh the checks.

CHANGELOG.md
```diff
@@ -17,7 +17,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Fixed -->
 <!-- ### Removed -->
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Make hidden code blocks readonly to prevent accidental erasure
```
AndrewJakubowicz, line 24:
```suggestion
- Make hidden code blocks readonly to prevent accidental erasure.
```

AndrewJakubowicz:
Checks are back! 🎉

AndrewJakubowicz:
Not urgent. Requesting re-review @aomarks. Thank you!

PR #282, opened 2022-03-16
[preview] expose iframe

e111077:
expose the iframe to enable load events and communication

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Fixed -->
 <!-- ### Removed -->
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- playgroun-preview now exposes the internal iframe via the `iframe` property.
 
 ## [0.15.0] - 2022-03-07
 
--- a/README.md
+++ b/README.md
@@ -788,11 +788,11 @@ A pure text editor based on CodeMirror with syntax highlighting for HTML, CSS, J
 ### Properties
 
 | Name       | Type                       | Default        | Description                                                                                                                    |
-| ---------- | -------------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| ---------- | -------------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------ |
 | `project`  | `string PlaygroundProject` | `undefined`    | The project that this editor is associated with. Either the `<playground-project>` node itself, or its `id` in the host scope. |
 | `location` | `string`                   | `""`           |                                                                                                                                |
 | `htmlFile` | `string`                   | `"index.html"` | The HTML file used in the preview.                                                                                             |
-| `iframe`   | `HTMLIFrameElement|null`   | `null`         | A reference to the internal iframe element that is used to render the preview.                                                 |
+| `iframe`   | `HTMLIFrameElement         | null`          | `null`                                                                                                                         | A reference to the internal iframe element that is used to render the preview. |
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -788,11 +788,11 @@ A pure text editor based on CodeMirror with syntax highlighting for HTML, CSS, J
 ### Properties
 
 | Name       | Type                       | Default        | Description                                                                                                                    |
-| ---------- | -------------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------ |
+| ---------- | -------------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------ |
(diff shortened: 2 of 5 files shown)


src/playground-preview.ts
```diff
@@ -107,8 +107,8 @@ export class PlaygroundPreview extends PlaygroundConnectedElement {
   @property()
   location = 'Result';
 
-  @query('iframe')
-  private _iframe!: HTMLIFrameElement;
+  @query('iframe', true)
+  iframe!: HTMLIFrameElement;
```
aomarks, line 111: Maybe we should remove the `!` since the user could call it before `render`.

aomarks (changes requested):
Mind adding a `CHANGELOG.md` entry, and also adding a row to the table here?: https://github.com/google/playground-elements#playground-preview

aomarks (changes requested):
Also, a test that confirms that the `iframe` object remains the same across preview reloads would be nice, because there's a risk in the future that we would change the way we manage iframes and break this assumption.

src/test/playground-ide_test.ts
```diff
@@ -649,6 +650,63 @@ suite('playground-ide', () => {
     assert.equal(historyLengthAfter, historyLengthBefore);
   });
 
+  test('reloading preview does not create a new iframe element', async () => {
+    // NOTE: For some reason, the parent window's history only seems to be
```
aomarks, line 654: Shouldn't need the separate origin in this test case, that's only for testing the history issue.

src/test/playground-ide_test.ts
```diff
@@ -649,6 +650,63 @@ suite('playground-ide', () => {
...
+      'playground-preview'
+    )) as PlaygroundPreview;
+
+    const iframe = (await pierce(
+      'playground-ide',
+      'playground-preview',
+      'iframe'
+    )) as HTMLIFrameElement;
+
+    assert.equal(preview.iframe, iframe);
```
aomarks, line 683: This assertion seems unnecessary.

src/test/playground-ide_test.ts
```diff
@@ -649,6 +650,63 @@ suite('playground-ide', () => {
...
+        </playground-ide>
+      `,
+      container
+    );
+    const preview = (await pierce(
+      'playground-ide',
+      'playground-preview'
+    )) as PlaygroundPreview;
+
+    const iframe = (await pierce(
```
aomarks, line 677: Could just do `iframe = preview.iframe`.

aomarks:
> Mind adding a CHANGELOG.md entry, and also adding a row to the table here?: https://github.com/google/playground-elements#playground-preview

ping

CHANGELOG.md
```diff
@@ -17,7 +17,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Fixed -->
 <!-- ### Removed -->
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- playgroun-preview now exposes the internal iframe via the `iframe` property.
```
aomarks, line 24:
```suggestion
- playground-preview now exposes the internal iframe via the `iframe` property.
```

README.md
```diff
@@ -788,10 +788,11 @@ A pure text editor based on CodeMirror with syntax highlighting for HTML, CSS, J
...
+| ---------- | -------------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------ |
 | `project`  | `string PlaygroundProject` | `undefined`    | The project that this editor is associated with. Either the `<playground-project>` node itself, or its `id` in the host scope. |
 | `location` | `string`                   | `""`           |                                                                                                                                |
 | `htmlFile` | `string`                   | `"index.html"` | The HTML file used in the preview.                                                                                             |
+| `iframe`   | `HTMLIFrameElement         | null`          | `null`                                                                                                                         | A reference to the internal iframe element that is used to render the preview. |
```
aomarks, line 795: The `|` here is being interpreted as a column separator, so it broke the table. I think you can escape it like:

```
`HTMLIFrameElement \| null`
```

aomarks (changes requested):

README.md
```diff
@@ -792,6 +792,7 @@ A pure text editor based on CodeMirror with syntax highlighting for HTML, CSS, J
 | `project`  | `string PlaygroundProject` | `undefined`    | The project that this editor is associated with. Either the `<playground-project>` node itself, or its `id` in the host scope. |
 | `location` | `string`                   | `""`           |                                                                                                                                |
 | `htmlFile` | `string`                   | `"index.html"` | The HTML file used in the preview.                                                                                             |
+| `iframe`   | `HTMLIFrameElement\|null`  | `null`         | A reference to the internal iframe element that is used to render the preview.                                                 |
```
aomarks, line 795:
```suggestion
| `iframe`   | `HTMLIFrameElement \| null`  | `null`         | A reference to the internal iframe element that is used to render the preview.                                                 |
```

Nit: We have spaces between all the other cases like this

aomarks (changes requested):

aomarks (approved):

PR #284, opened 2022-03-16
Release 0.15.1

e111077:
LMK if there's other stuff that needs to be done around the release


aomarks:
> LMK if there's other stuff that needs to be done around the release

You also need to run `npm run update-version-module` (see https://github.com/google/playground-elements/runs/5577996813?check_suite_focus=true)

aomarks (approved):
Also add a release to https://github.com/google/playground-elements/releases after publish

PR #287, opened 2022-03-18
feat: linewrapping for editor

jholt1:
Add codemirror functionality for wrapping long lines, with additional script to keep the indent.


src/playground-code-editor.ts
```diff
@@ -216,6 +216,13 @@ export class PlaygroundCodeEditor extends LitElement {
   @property({type: Boolean, attribute: 'line-numbers', reflect: true})
   lineNumbers = false;
 
+  /**
+   * If true, wrap for long lines. Default false
+   * (hidden).
```
aomarks, line 221: What does "hidden" mean here?

README.md
```diff
@@ -668,18 +668,19 @@ All-in-one project, editor, file switcher, and preview with a horizontal side-by
...
+| `projectSrc`         | `string`                         | `undefined`               | URL of the [project manifest](#project-manifest) to load                                                                                                                                          |
+| `config`             | `ProjectManifest`                | `undefined`               | Get or set the project configuration and files, ([details](#option-3-config-property)).                                                                                                           |
+| `lineNumbers`        | `boolean`                        | `false`                   | Render a gutter with line numbers in the editor                                                                                                                                                   |
+| `lineWrapping`       | `boolean`                        | `false`                   | Should editor scroll or wrap for long lines                                                                                                                                                       |
```
aomarks, line 676:
```suggestion
| `lineWrapping`       | `boolean`                        | `false`                   | If `true`, long lines are wrapped, otherwise the editor will scroll.                                                                                                                                                       |
```

src/playground-code-editor.ts
jholt1: Removed

jholt1:
Thanks for checking the PR @aomarks, I have fixed the formatting and added to the changelog

src/playground-code-editor.ts
```diff
@@ -495,6 +505,20 @@ export class PlaygroundCodeEditor extends LitElement {
       }
     });
 
+    cm.on('renderLine', (_editorInstance: Editor, line, elt) => {
+      if (!this.lineWrapping) {
```
aomarks, line 509: Sorry, for some reason I didn't notice this part of the code when I first looked at this PR.

Why do we need special handling here? Why doesn't the built-in `lineWrapping` option work as-is?

aomarks: Do you think we could add and remove the `renderLine` listener all together whenever `lineWrapping` changes inside `update`, instead of always listening and then aborting? Should be more efficient in the case where it is disabled.

src/playground-code-editor.ts
```diff
@@ -505,21 +506,31 @@ export class PlaygroundCodeEditor extends LitElement {
...
+    elt: HTMLElement
+  ) {
+    const basePadding = 4;
+    const charWidth = _editorInstance.defaultCharWidth();
+    const tabSize = _editorInstance.getOption('tabSize') || basePadding;
+    const off = CodeMirror.countColumn(line.text, null, tabSize) * charWidth;
+
+    elt.style.textIndent = `-${off}px`;
+    elt.style.paddingLeft = `${basePadding + off}px`;
   }
```
aomarks, line 534: `_` prefix on parameters usually means unused

```suggestion
  private _onRenderLine(
    editorInstance: Editor,
    line: CodeMirror.LineHandle,
    elt: HTMLElement
  ) {
    const basePadding = 4;
    const charWidth = editorInstance.defaultCharWidth();
    const tabSize = editorInstance.getOption('tabSize') || basePadding;
    const off = CodeMirror.countColumn(line.text, null, tabSize) * charWidth;

    elt.style.textIndent = `-${off}px`;
    elt.style.paddingLeft = `${basePadding + off}px`;
  }
```

src/playground-code-editor.ts
```diff
@@ -505,21 +506,31 @@ export class PlaygroundCodeEditor extends LitElement {
...
-      const tabSize = cm.getOption('tabSize') || basePadding;
-      const off = CodeMirror.countColumn(line.text, null, tabSize) * charWidth;
+    this._codemirror = cm;
+  }
 
-      elt.style.textIndent = `-${off}px`;
-      elt.style.paddingLeft = `${basePadding + off}px`;
-    });
+  private _indentingLine(cm: CodeMirror.Editor) {
+    if (this.lineWrapping) {
```
aomarks, line 515: I think I would actually just inline this directly in the `lineWrapping` case in `update`.

src/playground-code-editor.ts
```diff
@@ -505,21 +506,31 @@ export class PlaygroundCodeEditor extends LitElement {
...
+  }
 
-    this._codemirror = cm;
+  private _onRenderLine(
+    _editorInstance: Editor,
+    line: CodeMirror.LineHandle,
+    elt: HTMLElement
+  ) {
+    const basePadding = 4;
+    const charWidth = _editorInstance.defaultCharWidth();
```
aomarks, line 528: What if the font size is different? I guess the indentation will always be fixed, regardless of the actual font size?

src/playground-code-editor.ts
jholt1: The `lineWrapping` case only happens when the property changes. If it is initially set as true, then `renderLine` never runs unless it is also in the `_createView`.

aomarks: Sorry I mean this entire function could be removed, and both cases put inside the switch block. Just because the function is so short that it doesn't really seem to help readability.

src/playground-code-editor.ts
jholt1: Screenshot 2022-04-04 at 23 52 10
Screenshot 2022-04-04 at 23 48 58

Looks like it does okay with different fonts and font sizes. CodeMirror does have a note about defaultCharWidth.
> Note that for non-monospace fonts, this is mostly useless, and even for monospace fonts, non-ascii characters might have a different width

Doesn't seem too bad
image

aomarks: I wonder if we could use `em` or `ch` or a similar font-aware CSS unit? Then we might not need to ask codemirror for any font info at all.

https://developer.mozilla.org/en-US/docs/Web/CSS/length#syntax

jholt1: Using `ch` could most likely work for monospace fonts, which I am guessing most would be using a monospace font for the editor anyway. I can have a look into how this could be done.

I had a quick peek to see how CodeMirror was working it out https://github.com/codemirror/CodeMirror/blob/master/src/measurement/position_measurement.js#L607-L615 Looks like they try to get an average size and then cache it.

This is a nice article discussing the use of `ch` in a similar way https://meyerweb.com/eric/thoughts/2018/06/28/what-is-the-css-ch-unit/

aomarks: Cool, would you be able to try that out? Seems like it would be simpler, faster, and would dynamically adapt to the font.

jholt1: Yeah `ch` works better. I realised that there was an issue with the guttering as there is additional padding when line numbers are added. Due to the guttering needed to use `calc` for the padding left. Let me know what you think

src/playground-code-editor.ts
```diff
@@ -495,9 +510,31 @@ export class PlaygroundCodeEditor extends LitElement {
...
+
     this._codemirror = cm;
   }
 
+  private _onRenderLine(
+    editorInstance: Editor,
+    line: CodeMirror.LineHandle,
+    elt: HTMLElement
+  ) {
+    const basePadding = 4;
```
aomarks, line 525: Could you add a comment here about what the purpose of this style adjustment logic is? I'm not sure I actually understand what it's doing.

aomarks (changes requested):
Looking good, very sorry for the extremely long review process. I just have a question that could do with a comment -- but should be ready to merge after that.

src/playground-code-editor.ts
jholt1: @aomarks added a comment, hope it makes a little more sense now 😄

aomarks (approved):
Thanks again for the PR!

src/playground-code-editor.ts
aomarks: Yes, very helpful, thanks!

aomarks:
This is now published as 0.15.5

PR #289, opened 2022-03-23
Release 0.15.2

AndrewJakubowicz:
### Added

- Make hidden code blocks readonly to prevent accidental erasure.
- Added `Ctrl+/` or `Cmd+/` hotkey for toggling line comments.
- Added keyboard shortcut documentation to the playground-code-editor README.

PR #290, opened 2022-03-23
Fix iframe test to wait for load event

AndrewJakubowicz:
This should resolve the timeout error encountered in https://github.com/google/playground-elements/pull/289


aomarks (approved):
Seems fine, but curious why bumping the timeout isn't equivalent?

AndrewJakubowicz:
Thanks for the comment @aomarks. While investigating I think I've root caused the timeout.
Since this was a single test timeout, splitting it out isn't a valid fix. It was a naive attempt at a quick fix.

I believe the real fix is waiting for the iframe to load in the same way that is done in other tests. Updated and tested locally. Now investigating why Github actions is not running on this PR?

aomarks (approved):

PR #293, opened 2022-03-23
fix flaky test

e111077:
This test flakes and my guess is that it will do so awaiting the iframe load which just watches for the `load` event. It seems that the `save` method will do a compilation and then trigger a reload of the iframe within it. My guess is that we are adding the iframe load event listener too late in some flaky cases.

### DIFF
--- a/src/test/playground-ide_test.ts
+++ b/src/test/playground-ide_test.ts
@@ -684,8 +684,9 @@ suite('playground-ide', () => {
       'playground-project'
     )) as PlaygroundProject;
 
-    await project.save();
-    await waitForIframeLoad(iframe);
+    const savePromise = project.save();
+    const reloadPromise = waitForIframeLoad(iframe);
+    await Promise.all([savePromise, reloadPromise]);
 
     const newIframe = (await pierce(
       'playground-ide',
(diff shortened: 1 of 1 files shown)


src/test/playground-ide_test.ts
```diff
@@ -684,8 +684,9 @@ suite('playground-ide', () => {
       'playground-project'
     )) as PlaygroundProject;
 
-    await project.save();
-    await waitForIframeLoad(iframe);
+    const savePromise = project.save();
+    const reloadPromise = waitForIframeLoad(iframe);
+    await Promise.all([savePromise, reloadPromise]);
```
aomarks, line 689: Inlining seems a bit easier to read (fewer names to read):

```suggestion
    await Promise.all([waitForIframeLoad(iframe), project.save()]);
```

aomarks (approved):
Nice catch

e111077:
nice catch or more like nice bug-writing?

AndrewJakubowicz (changes requested):
Locally I patched this in and updated `export const npmVersion = '0.15.2';` in `version.ts` and the timeout reproduced :(

AndrewJakubowicz:
Will circle back in a second. Running a clean test with version bump (nuking npm).

AndrewJakubowicz (approved):

e111077:
we have another flaky test, but it doesn't seem to be related to this one. So imma go ahead and merge

PR #295, opened 2022-03-25
Fix and cleanup fold and hide logic

AndrewJakubowicz:
Fixes readonly characters when code folding (commit 1).
Fixes https://github.com/google/playground-elements/issues/267 (commit 2).

Note to reviewer:
 - Commit 1 fixes the `hide` folding of marker comments so they're not readonly.
 - Commit 2 replaces the TextMarker clearing logic such that it doesn't add history.
 - Commit 3 is changelog.

### Context

Playground elements has a great [Hiding & folding](https://github.com/google/playground-elements#hiding--folding) feature that uses comments to mark regions of code that can be folded or hidden. Regardless of whether hiding or folding, the comment markers themselves are always hidden. E.g.

```ts
/* playground-fold */
console.log('hi');
/* playground-fold-end */
```

Results in CodeMirror markers which can be visualized as:

```
hide(/* playground-fold */)
fold(console.log('hi');)
hide(/* playground-fold-end */)
```

A recent change causes all `hide` regions to be readonly. In the `fold` case this results in the folded region still being editable, but the comment markers creating invisible markers that cannot be deleted.

### Fix

Make the comment markers not readonly. Thus when used as part of either folding of hiding they can be deleted. The deletion of a hidden region is atomic so the marker can be completely deleted by a user in one keystroke. Prior behavior was that the delete key would not delete the comment marker and cursor would get stuck (if pressing delete multiple times).

### Investigation that lead to second commit

I also investigated trying to clear the hide/fold comments from the document value, however any change to the codeMirror value resulted in history pollution.
While investigating this I also inspected the way we reset all the text markers by clearing and resetting the document contents.

I noticed the marker clearing technique also added history state by pushing an empty string onto the history. By refactoring it to clear the markers we no longer pollute history and can simplify some state variables.

### Testing

I tested both folding and hiding in the configurator, and also fixed the broken unit test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Fixed -->
 <!-- ### Removed -->
 
+## Unreleased
+
+### Fixed
+
+- Fixed the hiding of comments that surround a fold/hide code block from
+  creating invisible readonly regions.
+
+- Refactored fold/hide marker logic so it doesn't add document history when
+  clearing the prior fold/hide markers.
+
 ## [0.15.2] - 2022-03-24
 
 ### Added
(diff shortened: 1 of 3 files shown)


aomarks (approved):
Nice

Issue #303, opened 2022-04-04
Add ability to focus a specific file when loading project


kevinpschaaf:
Proposal: Add a flag in the `project.json` `files` options to indicate a specific file in the project should be focused when a project is loaded.

e.g.
```json
{
  "files": {
    "index.html": {},
    "foo.ts": { "selected": true },
    "bar.ts": {}
  }
}
```

This is useful when using `playground-elements` in a tutorial flow, in order to direct a user to edit a specific file in the project.

AndrewJakubowicz:
Maybe unrelated - noting as may be relevant to this investigation. In the tutorial flow if a file is added and focused. If the user then goes to a prior part where that file doesn't exist, it remains open but with no content. This may be out of scope of playground-elements.

PR #304, opened 2022-04-05
Add ability to select a file in project.json

kevinpschaaf:
The selected file in the file tabs is selected based on the selected file in the project config only when loading a new project.

Fixes https://github.com/google/playground-elements/issues/303

### DIFF
--- a/src/playground-project.ts
+++ b/src/playground-project.ts
@@ -353,7 +353,9 @@ export class PlaygroundProject extends LitElement {
     this._pristineFiles =
       this._files && JSON.parse(JSON.stringify(this._files));
     this._modified = false;
-    this.dispatchEvent(new CustomEvent('filesChanged'));
+    this.dispatchEvent(
+      new CustomEvent('filesChanged', {detail: {projectLoaded: true}})
+    );
     this.save();
   }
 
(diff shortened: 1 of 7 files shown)


src/playground-project.ts
```diff
@@ -51,6 +51,20 @@ const generateUniqueSessionId = (): string => {
...
 };
 
+declare global {
+  interface HTMLElementEventMap {
+    filesChanged: FilesChangedEvent;
+  }
+}
+
+export class FilesChangedEvent extends Event {
+  projectLoaded: boolean;
```
aomarks, line 61: WDYT about a separate `projectLoaded` event? Seems like that could be a little simpler.

aomarks (changes requested):
Looks good, just a thought about whether we should have 2 events instead of a bit.

Also, could you add a CHANGELOG entry, and update these two sections of the README to add `selected`?

https://github.com/google/playground-elements#option-1-inline-scripts
https://github.com/google/playground-elements#option-2-json-configuration

aomarks:
Also looks like it needs an `npm run format`

aomarks (approved):

src/playground-project.ts
kevinpschaaf: The current work of `filesChanged` needs to happen for both "project loaded" and "project files changed" cases, so it'd be a bit more complex to listen both events and dedupe them correctly. Leaving as is per discussion.

AndrewJakubowicz (approved):
Superb!

This also provides a great way to handle the removal of files between lit.dev tutorial steps.

PR #313, opened 2022-05-11
Add browser to export conditions and release 0.16.0

aomarks:
Adds `browser` to the list of export conditions we use for resolving modules. This condition is listed in the "community conditions definitions" in the Node docs at https://nodejs.org/api/packages.html#community-conditions-definitions. We encountered it in a package we are using on lit.dev: https://unpkg.com/@floating-ui/dom@0.5.0/package.json

cc @sorvell


src/test/bare-module-worker_test.ts
```diff
@@ -1039,6 +1039,88 @@ suite('bare module worker', () => {
...
+                      "import": {
+                        "browser": "./browser/index.js"
+                      },
+                      "default": "./default/index.js"
+                    },
+                    "./bar.js": {
+                      "import": {
+                        "browser": "./browser/bar.js"
+                      },
+                      "default": "./default/bar.js"
```
justinfagnani, line 1083: what about exports without `"import"`? ie, in Lit we use default and development, conceivably one of them could have a browser variant.

aomarks: There are tests above this one that cover single export conditions, like just "imports" and just "development". I think given the implementation (an unordered set of conditions), we don't need to test all permutations (i.e. just "development" should work exactly like just browser").

justinfagnani (approved):

src/test/bare-module-worker_test.ts
```diff
@@ -1039,6 +1039,88 @@ suite('bare module worker', () => {
...
+              },
+              'default/bar.js': {
+                content: 'DEFAULT',
+              },
+              'package.json': {
+                content: `{
+                  "exports": {
+                    ".": {
+                      "import": {
+                        "browser": "./browser/index.js"
```
AndrewJakubowicz, line 1075: Worth adding a test matching the structure for the specific example?

https://unpkg.com/@floating-ui/dom@0.5.0/package.json

Example containing: `import.browser.default : <filename>`

aomarks: I think the test coverage is already ok, there are some other tests in here handling defaults. We also use the actual implementation from Node itself so the risk is fairly low to begin with.

PR #314, opened 2022-05-16
Allow consumer to use any/latest version mwc-elements

aarondrabeck:
Fixes the famous:

```my-app.ts:118 DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "mwc-ripple" has already been used with this registry```

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,12 +85,12 @@
     "typescript": "^4.4.3"
   },
   "dependencies": {
-    "@material/mwc-button": "^0.25.1",
-    "@material/mwc-icon-button": "^0.25.1",
-    "@material/mwc-linear-progress": "^0.25.1",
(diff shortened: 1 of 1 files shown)


aomarks:
Thanks for the PR, however, we can't do this because it would mean people who install MWC will always get the latest version of MWC by default, which could have a breaking change in it that is incompatible with playground-elements.

We should update the latest version of MWC if we are behind, but we can't just assume there will never be a breaking change, because they happen often.

aarondrabeck:
Sounds good, should I submit a PR for package bump?

aomarks:
> Sounds good, should I submit a PR for package bump?

That would be great! Make sure to run the tests if you can. It would also be great to read through the MWC changelog to see if any breaking changes are reported.

PR #317, opened 2022-05-26
Convert playground-elements to Wireit

aomarks:
- Converts everything to Wireit
- Adds a raf defer to make a test a little less flakey.
- Separates all of the rollup scripts so that they can run in parallel and more incrementally.
- Consolidates the format and lint scripts into one GitHub action workflow.


.github/workflows/lint-and-format.yml
```diff
@@ -12,6 +12,8 @@ jobs:
         with:
           node-version: 16
           cache: npm
+      - uses: google/wireit@setup-github-actions-caching/v1
 
       - run: npm ci
+      - run: npm run lint
```
rictic, line 18: Consider adding a lint-and-format build script

or just make `npm run lint` run both in parallel, as format checking feels like a linting

PR #323, opened 2022-07-06
Code editor jsx / tsx syntax highlight

taylor-vann:
- Add `jsx` support to codemirror
- `tsx` highlighting is included by default in codemirror `jsx` support

This change adds some KB to the bundle:
+ 24.64 KB uncompressed uniminified
+ 6.83 KB Brottli


src/playground-file-editor.ts
```diff
@@ -226,6 +226,11 @@ const mimeTypeToTypeEnum = (mimeType?: string) => {
     case 'text/javascript':
     case 'application/javascript':
       return 'js';
+    case 'text/jsx':
+      return 'jsx';
+    // tsx uses the same syntax highlighting as JSX
+    case 'text/typescript-jsx':
```
aomarks, line 232: I think we should have a distinct `jsx` vs `tsx` filetypes. The fact that both filetypes are supported by the same codemirror syntax highlighter is an implementation detail, so that can just be represented in `getLanguageMode`.

So I think this case should return `tsx`.

src/playground-code-editor.ts
```diff
@@ -207,7 +207,7 @@ export class PlaygroundCodeEditor extends LitElement {
    * extension.
    */
   @property()
-  type: 'js' | 'ts' | 'html' | 'css' | 'json' | undefined;
+  type: 'js' | 'ts' | 'html' | 'css' | 'json' | 'jsx' | undefined;
```
aomarks, line 210: Add `tsx` (see other comment below)

src/playground-code-editor.ts
```diff
@@ -931,6 +931,7 @@ export class PlaygroundCodeEditor extends LitElement {
       case 'js':
       case 'ts':
       case 'css':
+      case 'jsx':
```
aomarks, line 934: Add `tsx` (see other comment)

src/playground-code-editor.ts
```diff
@@ -960,6 +961,8 @@ export class PlaygroundCodeEditor extends LitElement {
         return 'google-html';
       case 'css':
         return 'css';
+      case 'jsx':
```
aomarks, line 964: Add `tsx` which should return `jsx` here (see other comment)

aomarks (changes requested):
Nice!

Please add 2 test cases to https://github.com/google/playground-elements/blob/545721fbf75f633823c20b5affe749cc846ddc9b/src/test/playground-code-editor_test.ts#L271 -- one for `jsx` and one for `tsx`.

src/playground-code-editor.ts
taylor-vann: Done!

src/playground-code-editor.ts
taylor-vann: sg done!

src/playground-code-editor.ts
taylor-vann: sg done!

taylor-vann:
sweet! added a couple tests and addressed comments :)

aomarks (changes requested):
Looks great!

Would you mind updating the PR description to include the difference in size between the old and new versions of `codemirror-bundle.js`?

taylor-vann:
sounds good! added how much comes with the new modes:

```js
// + 24.64 KB uncompressed uniminified
// + 6.83 KB Brottli
```

taylor-vann:
figured the size might change but the delta change for their addition is what's important

taylor-vann:
@AndrewJakubowicz I believe you're writting TS in JSX `declare` is a `tsx` thang :)

taylor-vann:
lemme boot and test manually again real quick. Might have biffed the tsx highlighting with those extra case statements. Stuff worked before

aomarks (approved):
Looks good! Let's just make sure we understand why it's not working for @AndrewJakubowicz

AndrewJakubowicz:
@taylor-vann and I paired.

On initial load the `.tsx` file is sent down with an unexpected mimetype.

For example:
```
{name: 'my-element.tsx', content: '\ndeclare const React;\n\nconst thing1 = () => <p>Hello <b>{101}</b>!</p>;\n', contentType: 'application/octet-stream'}
```

However when the file is dynamically added - it is correctly handled.

aomarks:
> @taylor-vann and I paired.
> 
> On initial load the `.tsx` file is sent down with an unexpected mimetype.
> 
> For example:
> 
> ```
> {name: 'my-element.tsx', content: '\ndeclare const React;\n\nconst thing1 = () => <p>Hello <b>{101}</b>!</p>;\n', contentType: 'application/octet-stream'}
> ```
> 
> However when the file is dynamically added - it is correctly handled. Renaming the file also fixes the issue on rename.

Ah, I think we prefer the mime type from the server itself, before we infer it from the file extension. Web dev server presumably isn't setting the mime type we are expecting, so it ends up with unknown type.

I kind of think we should drop mime types all together and just rely purely on file extensions. Servers don't sniff mime types consistently.

aomarks:
> > @taylor-vann and I paired.
> > On initial load the `.tsx` file is sent down with an unexpected mimetype.
> > For example:
> > ```
> > {name: 'my-element.tsx', content: '\ndeclare const React;\n\nconst thing1 = () => <p>Hello <b>{101}</b>!</p>;\n', contentType: 'application/octet-stream'}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However when the file is dynamically added - it is correctly handled. Renaming the file also fixes the issue on rename.
> 
> Ah, I think we prefer the mime type from the server itself, before we infer it from the file extension. Web dev server presumably isn't setting the mime type we are expecting, so it ends up with unknown type.
> 
> I kind of think we should drop mime types all together and just rely purely on file extensions. Servers don't sniff mime types consistently.

I would be fine with just filing an issue about this and not blocking this PR. Also note you can manually specify the mime type in the project.json file, so there is a workaround for when the server sends the wrong mime type.

configurator/project/project.json
```diff
@@ -1,5 +1,6 @@
 {
   "files": {
+    "howdy.tsx": {},
```
AndrewJakubowicz, line 3: Nit: remove

taylor-vann:
> I kind of think we should drop mime types all together and just rely purely on file extensions. Servers don't sniff mime types consistently.

agree

src/playground-project.ts
```diff
@@ -801,12 +801,17 @@ const expandProjectConfig = async (
       filePromises.push(
         (async () => {
           const resp = await fetch(new URL(filename, baseUrl).href);
+          let contentType = resp.headers.get('Content-Type')?.toLowerCase() ?? 'text/plain';
+          // native fetch api will default to octet-stream for unknown filetypes
```
aomarks, line 805: I think it'll be the server doing this, not fetch.

The file extension fallback logic already exists elsewhere, so no need to set the mime type here. But treating octet-stream the same as undefined seems like it makes sense.

aomarks (changes requested):
See comment about mime type logic.

src/playground-project.ts
taylor-vann: feel the same way, `application/mime-type` is historically a bit bucket

configurator/project/project.json
taylor-vann: done!

taylor-vann:
kewl rolled back the mime-type injection.

demo files can explicitly declare mimetypes in their pacakge.json files

src/playground-project.ts
taylor-vann: removed the injection

AndrewJakubowicz (approved):
Thank you!

src/playground-file-editor.ts
```diff
@@ -226,6 +226,11 @@ const mimeTypeToTypeEnum = (mimeType?: string) => {
     case 'text/javascript':
     case 'application/javascript':
       return 'js';
+    case 'text/jsx':
+      return 'jsx';
+    // tsx uses the same syntax highlighting as JSX
```
aomarks, line 231: This comment a bit confusing here. Belongs in the code-editor highlight  mode choosing function if anywhere.

aomarks (changes requested):
> I would be fine with just filing an issue about this and not blocking this PR. Also note you can manually specify the mime type in the project.json file, so there is a workaround for when the server sends the wrong mime type.

Mind also filing an issue to track this? I think ignoring server mime type is what we should probably do in a follow-up sometime.

src/playground-file-editor.ts
taylor-vann: im just gonna remove it

taylor-vann:
@aomarks created https://github.com/google/playground-elements/issues/329 to come back to

src/playground-file-editor.ts
taylor-vann: done!

aomarks (approved):

PR #324, opened 2022-07-18
[infra] Update wireit and fix output config

augustjk:
I noticed the `build` command was failing due to overlapping output of `internal/typescript.js` which was getting erroneously cleaned out.

Updating `wireit` to the latest version with output modified detection surfaced additional overlaps.


aomarks:
> @aomarks Also I don't have write access for this repo so can't run the workflows or merge. pray

Sent invite

PR #333, opened 2022-08-17
Modifications for Google import compatibility

aomarks:
Various changes to make it easier to import playground elements into Google:

- Remove `import type` statements.
- Require explicit `override` keywords.
- Replace `vscode-languageserver` dep with `vscode-languageserver-protocol`.
- Detect and annotate floating promises, so that they can be auto transformed to have the required annotation internally.
- Upgrade from TypeScript 4.4.4 to 4.7.4, and use `~` semver range going forward.
- Add explicit cast for a `JSON.parse` call.
- Add explicit type parameters for a `new Map` call.
- Rename variables called `exports` to `_exports`, because JS Compiler assumes it can always unambiguously refer to a global with that name.
- Simplify part of the project -> service worker proxy protocol. We used to pass the service worker URL over the wire, but since it's static there is no point doing that (simplifies some internal security transforms).

### DIFF
--- a/src/test/typescript-worker_test.ts
+++ b/src/test/typescript-worker_test.ts
@@ -6,8 +6,8 @@
 
 import {checkTransform} from './worker-test-util.js';
 
-import type {BuildOutput, SampleFile} from '../shared/worker-api.js';
-import type {CdnData} from './fake-cdn-plugin.js';
+import {BuildOutput, SampleFile} from '../shared/worker-api.js';
+import {CdnData} from './fake-cdn-plugin.js';
 
 suite('typescript builder', () => {
   test('empty project', async () => {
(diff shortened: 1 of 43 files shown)


src/test/typescript-worker_test.ts
```diff
@@ -279,7 +279,7 @@ suite('typescript builder', () => {
               },
               'index.d.ts': {
                 content: `
-                  import type {t} from './type.js';
+                  import {t} from './type.js';
```
rictic, line 282: this doesn't need to change

rictic (approved):

src/test/typescript-worker_test.ts
aomarks: Removed

PR #336, opened 2022-09-15
Skip setting file to language service if exists and has not changed

Matsuuu:
A small bug that had slipped through the reviews in type language service rework.

Old implementation ran `this.files.set` every time a file was found, but wasn't modified. To reduce unnecessary map access, we just early return if..

- A file is already in the language service host
- But isn't updated.


Matsuuu:
If https://github.com/google/playground-elements/pull/337 gets merged, this can be discarded since that PR also contains this fix

aomarks (approved):

PR #338, opened 2022-10-06
Add require-corp embedder policy header to service worker

AndrewJakubowicz:
Upstream a recommended header that enforces CORS (as far as I can tell).
Reference to documentation: https://html.spec.whatwg.org/multipage/origin.html#coep

I tested this using the configurator page and a relatively complex demo. However I'm not certain that this is an appropriate test for a modification to the service worker.

https://user-images.githubusercontent.com/15080861/194410336-8bb12c31-fa45-469d-bdf0-7d8f6ad95901.mov


AndrewJakubowicz:
On further testing this change would be breaking. For example, we often use resources from another origin for demonstrating example images. Any image elements without the `crossorigin` attribute would break. E.g.:

Screen Shot 2022-10-10 at 10 34 58 AM

Two proposed options:
 - Make this opt-in (if possible).
 - Revert this change and keep this as an internal only change.

PR #347, opened 2022-11-10
Upgrade TypeScript and everything else, and switch to nodenext and es2021

aomarks:
- Upgrade TypeScript, codemirror, and everything else too.

- Switches to `moduleResolution: nodenext`. This makes TypeScript follow `export` conditions. For Lit, that means we'll now see an error when the `.js` extension is missing from an import.

- Because the `type` and `imports/exports` fields are now significant, we now pass the `package.json` to TypeScript. In addition, because the default `type` is `commonjs`, we force the default to be `module` instead, so that we get module semantics by default.

- Switches `target` from `es2017` to `es2021`, which is supported in all modern browsers.

- Upgrades to the latest Wireit to get services, and configured for the `serve` command.


justinfagnani (approved):

src/test/completions_test.ts
```diff
@@ -29,7 +29,7 @@ suite('completions', () => {
           <script type="sample/ts" filename="hello.ts"></script>
           <script type="sample/html" filename="index.html">
             <body>
-              <script src="hello.js">&lt;/script>
+              <script type="module" src="hello.js" type="module">&lt;/script>
```
justinfagnani, line 32: This has two type=module attributes

PR #369, opened 2023-05-24
Gracefully handle a TypeScript compiler crash once

AndrewJakubowicz:
### Context

This fixes https://github.com/lit/lit.dev/issues/1128 by providing a path to gracefully recover from bad incremental state.

### Testing

Manually confirmed this fix works - investigating adding unit testing.

Note that for the specific case that this PR was created, it's also viable to upgrade the bundled TypeScript which resolves the compiler error. A benefit of this solution is we could use it for other rare exceptions in the future.

### DIFF
--- a/src/internal/build.ts
+++ b/src/internal/build.ts
@@ -17,7 +17,7 @@ import {Diagnostic} from 'vscode-languageserver-protocol';
 
 const unreachable = (n: never) => n;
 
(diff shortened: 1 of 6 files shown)


AndrewJakubowicz:
This is not needed since we upgraded TypeScript to 5.

PR #375, opened 2023-08-10
Fix tab to spaces replacement logic

### LINKED ISSUE #358: Tab always inserts 1 space instead of 2 using default options
As in my comment in https://github.com/google/playground-elements/pull/129, because of an error in the code, the number of spaces inserted is always one fewer than expected (with default settings, it means 1 instead of 2).

If I can get the CLA figured out I can create a pull request, but otherwise anyone interested can take this over. (Solution is already in the link)

MichaelByrneAU:
The existing implementation of converting a tab into spaces inadvertently includes an off-by-one error, producing one less space than desired. Resolves #358.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix tab-to-spaces conversion in the code editor which was resulting in one less space being produced on tab than desired.
+
 ## [0.18.0] - 2023-05-25
 
 ### Changed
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz (commented):
Thank you – looks awesome.

Could you also include a small changeset entry? E.g. an unreleased bullet point under fixed. Something like https://github.com/google/playground-elements/pull/368/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed.

Thank you!

MichaelByrneAU:
Done!

AndrewJakubowicz (approved):

PR #398, opened 2024-03-20
Speed up builds by not bundling typescript

justinfagnani:
This marks `internal/typescript.js` as internal in the worker Rollup config, and switches the worker to use modules so it can load the typescript module build.

`playground-typescript-worker.js`'s size goes from 3034427 bytes to 39289 bytes, and the build time on my M1 Pro goes from 26.2s to 4.4s when the typescript bundle is already built.

Module worker support is very good now, at > 96% of tracked users: https://caniuse.com/mdn-api_worker_worker_ecmascript_modules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Upgraded to TypeScript 5.2 and Lit 3.0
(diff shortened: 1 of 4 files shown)


aomarks (approved):
🎉

tbroyer:
This introduced an (unexpected?) breaking change: the `internal/typescript.js` now needs to be copied alongside the `playground-typescript-worker.js`, and when using Rollup with `@web/rollup-plugin-import-meta-assets`, it has to be manually copied to the `assets` folder (alongside the worker script that Rollup would have copied itself).

The README needs to be updated to reflect this. See #418

PR #410, opened 2024-08-15
Update runtime and test dependencies

justinfagnani:
New versions of Fuse.js and tslib play better with modules since they define exports for JS and types.


aomarks (approved):

justinfagnani:
I can't see from the logs what's causing Firefox to fail on CI, and tests pass locally for me. I'm going to try to update the testing deps and see if that helps.

aomarks (approved):

aomarks:
> @aomarks looks like tests pass with going all-in on Playwright! Do you want a separate PR for updating the test deps, or is this one ok?

One seems good

PR #422, opened 2025-03-27
feat(cdn): allows customization of the CDN base url

e111077:
Allows for setting the CDN base url. Currently only tested is unpkg and https://cdn.jsdelivr.net/npm via the `cdnBaseUrl` property and config entry

### DIFF
--- a/src/configurator/playground-configurator.ts
+++ b/src/configurator/playground-configurator.ts
@@ -130,7 +130,8 @@ export class PlaygroundConfigurator extends LitElement {
         flex: 1;
       }
 
-      .knobs select {
+      .knobs select,
+      .knobs input[type='text'] {
         grid-column: 2 / -1;
       }
 
@@ -163,7 +164,7 @@ export class PlaygroundConfigurator extends LitElement {
     `,
   ];
 
-  private values = new KnobValues();
+  private _values = new KnobValues();
 
   @state()
   private _themeDetectorOpen = false;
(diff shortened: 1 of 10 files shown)


e111077 (commented):
Had to change the package.json because unpkg can't serve 2.8.0 for some reason but it can 3.0.0 so this unbreaks the default case

sorvell (approved):
Thanks so much! optional nits.

e111077:
Test failure seems like a flake

src/configurator/playground-configurator.ts
```diff
@@ -561,6 +572,21 @@ ${props.join('\n')}
     `;
   }
 
+  private inputKnob(knob: KnobsOfType<'input'>) {
```
justinfagnani, line 575: Nit: Private field names should have a _ prefix. But a big deal because this is the configurator.

justinfagnani (approved):

src/configurator/playground-configurator.ts
e111077: changed all private props and methods in file to have `_` prefix

src/playground-project.ts
```diff
@@ -174,6 +174,16 @@ export class PlaygroundProject extends LitElement {
...
 
+  /**
+   * Base URL for the CDN used to resolve bare module specifiers.
+   *
+   * Examples:
+   * - "https://unpkg.com" (default)
+   * - "https://cdn.jsdelivr.net/npm"
+   */
+  @property({attribute: 'cdn-base-url'})
+  cdnBaseUrl = 'https://unpkg.com';
```
augustjk, line 185: do we need this as a separate property on `<playground-project>` if it's already part of the `config` property, or the JSON in `projectSrc`?

e111077: The idea was that the property takes precedence over the config and if neither are defined then it defaults to unpkg.com. But actually what you're pointing out right now as well is that by setting this default on the property, it will by default not read from the config unless you explicitly set `''`

I'll work on a lower priority followup to this

e111077: #423 but I think your original question is still valid. It's a decision I made to support both. My reasoning was that if we wanted to set this across lit.dev, we'd only have to change the base config that is shared across basically all the projects

augustjk: yeah keeping it in the JSON manifest for the `project-src` or `config` property which share the same interface makes sense, and that'll allow us to update that for the whole of lit.dev playground in relatively few places since we extend from just a few base project jsons.

adding the extra `cdnBaseUrl` property on top of that to `<playground-project>` that overrides the project manifest doesn't seem as necessary to me and we'd have to clearly document which value takes precedence if we do keep it.

